Issue #1149, opened 2014-01-22
Multiple document shown in 0.6.0 when initializing documents pressing Ctrl+D multiple times in Ubuntu


albusshin:
To reproduce:
Open a instance of Light Table and open a Instarepl quickly.
Type the code of `(print 'x)` and then click on `print` and press `Ctrl` + `D` three times.
After the REPL is connected, there will be multiple docs shown in the editor, and we can only close one of them pressing `Ctrl` + `D` again.

PR #1168, opened 2014-01-25
Possible fix for issue #1149 - Multiple document shown when initializing documents pressing Ctrl+D multiple times

Morgawr:
First off, as I also wrote https://github.com/LightTable/LightTable/pull/1166 in this pull request, I am committing on my version of master that is a bit ahead (sorry I screwed up branching), let me know if it's a problem.

The actual core of the matter is the issue of multiple docstrings when a live repl is connecting. When it's connecting a message is sent with :try-send! that either queues the message if it's not connected or delivers it straight to the server. The problem is that queuing the same message (in case of docstring request) multiple times delivered it multiple times as well and the client would then get hammered with multiple docstrings.

This is a simple "quick and dirty" fix that makes sure :queue! operation is idempotent and no multiple instances of the same message can be queued together. 

A better fix would either be to a) add a :send-once!/:queue-once! flag if this behavior can't be intended as default or b) actually make sure that the documentation shown in the editor at a specific line is shown only once and is unique so multiple docs don't conflict with each other at the same line.

Solution B is probably the best but I couldn't figure out how to implement this. Merge this only if you think it might be useful, it might be more dangerous and break other behaviors than anything, though.


kenny-evitt:
This is intended to cover #1149 and #1248.

@Morgawr I'm going to close this. Re-open (or comment to that effect) if you still think your changes should be merged.

Issue #972, opened 2014-01-08
Debian 7 GLIBC


siscia:
After all the installation process on a debian 7 64bit I get
```shell
simo@simo:~/sandBox/light-table-core-2/deploy$ ./ltbin 
./ltbin: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.14' not found (required by ./ltbin)
./ltbin: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.15' not found (required by ./ltbin)
```
it is a know problem #164 back in the day was fixed compiling with gcc 4.4, now I don't know how to manage it...

kenny-evitt:
This is *probably* irrelevant now that we've (a while ago now) merged the changes to migrate from node-webkit to { Atom Shell / Electron }.

PR #1302, opened 2014-02-20
Proposal for issue #972

pelmato:
Hi,

I made some changes in the linux build process. The goal was to make it independent from an older compiled version of LightTable. It allows to choose the node-webkit version and thus will solve the issue number #972. As I'm not sure about how the whole build process works, I may have forgotten some key points, so please let me know if I can improve anything.

Here are more details about the modifications I made :
- Don't download pre-compiled version of LightTable. Instead :
- Download node-webkit binaries and copy files into "deploy" directory,
- Create a "resources/default-plugins.tar.gz" file containing default plugins (copied from http://d35ac8ww5dfjyg.cloudfront.net/playground/bins/0.6.0/LightTableLinux64.tar.gz )
- Change binary version in "version.json" file.

I successfully tested this corrections on both Debian Linux Wheezy 32 and 64bit. Hope it can help to fix issue #972.

Any comment welcome !


BenjaminVanRyseghem:
Is this still a thing as we are dropping node-webkit?

pelmato:
Well, probably not. Hope the Linux version will be more stable.

cldwalker:
@BenjaminVanRyseghem Good point. No longer valid with electron

PR #1434, opened 2014-04-20
Add reasonable default keybindings for zooming

cldwalker:
These keybindings are a standard way of zooming in OSX
browsers.

### DIFF
--- a/deploy/settings/default/default.keymap
+++ b/deploy/settings/default/default.keymap
@@ -5,7 +5,9 @@
            "pmeta-shift-f" [:searcher.show]
            "ctrl-shift-d" [:docs.search.show]
            "pmeta-n" [:new-file]
-           "ctrl-space" [:show-commandbar-transient]}
+           "ctrl-space" [:show-commandbar-transient]
+           "pmeta-shift-=" [:window.zoom-in]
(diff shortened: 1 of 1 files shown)


thewhitetulip:
1. toggle comment with ctrl-/ 
This is done in almost all the code editors including but not limited to Eclipse, atom, sql developer, I think mysqlworkbench I am not sure though, it is a standard way to toggle comments, it makes life easier really.

2. toggle workspace with ctrl-\
This isn't a standard toggle workspace keybinding of any editor that i am aware of till date except Atom, the thing is normally in old code editors you didn't toggle the workspaces but here you might need to, while coding django I create two tabsets, one with the code and another with the browser to see the o/p inline

3. add browser tab with ctrl-b
This is a new binding and it truly is lighttable specific

4. add tabset with ctrl-t
5. remove tabset with shift-ctrl-t

these two are new bindings as well

6. using alt-[0-9] to switch between tabs rather than ctrl-[0-9]
This I am pretty sure all apps do, I am not sure in the Windows/Mac, but in the Linux world all the tabs can be navigated with alt-[0-9] (in firefox, chrome, terminal, geany, I think in eclipse etc)

7. toggle console by alt-c
this is a totally new binding, but a useful at that

PR #1458, opened 2014-05-07
External Browser: Try a default port for socket.io first

jakubholynet:
Try to use a hard-coded port, 5678, and only ask
for a random free port when it is not available.
Users that have only one connection thus get by
most of the time without having to change the port
in their HTML whenever LT restarts.

I think it is also better to log errors than to
swallow them.
I am, of course, open to discussions.

### DIFF
--- a/src/lt/objs/clients/ws.cljs
+++ b/src/lt/objs/clients/ws.cljs
@@ -54,8 +54,14 @@
 
 (def server
   (try
-    (let [ ws (.listen io 0)]
+    (let [ ws (.listen io 5678)]
       (.set ws "log level" 1)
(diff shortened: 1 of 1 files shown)


tclev:
I'd like to bump this issue if possible. Unless there is another way that I don't know about, it's cumbersome to have always change my html every time I restart LightTable.

cldwalker:
@kenny-evitt @rundis This looks reasonable to me. I'll QA tonight. I'll be merging tomorrow unless there's a concern

cldwalker:
@jakubholynet Thanks for the PR! Better late than never

jakubholynet:
<3

PR #1507, opened 2014-06-05
Add support for connecting to browsers via XHR

rigdern:
Some HTML-based app environments (e.g. Windows Web Apps) do not support
cross-domain script inclusion but they do support cross-domain XHRs. This
change builds support into LightTable to be able to connect to such app
environments to evaluate HTML, JavaScript, and CSS.

### DIFF
--- a/deploy/core/node_modules/lighttable/ws.js
+++ b/deploy/core/node_modules/lighttable/ws.js
@@ -43,7 +43,9 @@
   }
 
   var thisScript = document.getElementById("lt_ws");
-  var parts = thisScript.src.split(":");
+  var useXhr = thisScript.hasAttribute("data-xhr-src");
+  var src = useXhr ? thisScript.getAttribute("data-xhr-src") : thisScript.src;
+  var parts = src.split(":");
   if(parts.length > 2) {
     var host = parts[0]+":"+parts[1];
     var port = parts[2].split("/")[0];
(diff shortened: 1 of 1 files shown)


joshuafcole:
Hey Adam,
Sorry for the delayed response.
After investigating this, I think it might be best to extract this into a
plugin. Since it targets specific use cases, it may be confusing to users
to see the option in all cases. By making it a plugin, users will opt into
if if the have they are working in an environment where it's applicable. If
you need any help doing so, just let me know. Additionally, I'll leave this
open in case any of the other team members or community disagree so we can
discuss it further.

Cheers,
Josh

On Thu, Dec 18, 2014 at 10:46 AM, Adam Comella <notifications@github.com>
wrote:

> @joshuafcole <https://github.com/joshuafcole> @kenny-evitt
> <https://github.com/kenny-evitt> Let me know if there's anything else I
> can help with. As I stated above, I think you should be able to test the
> cross-domain XHRs in Safari as long as you are visiting a file:// URL.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/LightTable/LightTable/pull/1507#issuecomment-67534097>
> .
>



-- 
Screw the environment. Please print this email immediately. And then burn
it.

rigdern:
LightTable/Javascript#9 affects the UI and I agree that could be done as a plugin. This PR (LightTable/LightTable#1507) doesn't affect the UI and I want to double check that you think it should be done in a plugin too.

This PR updates the websocket code to support downloading code via XHR (currently it only downloads it via script tags). To do this as a plugin, I would probably have to copy and paste my version of LightTable's [ws.js](https://github.com/rigdern/LightTable/blob/c6f25af515161d7ed4a13ba50cf095d6378003c7/deploy/core/node_modules/lighttable/ws.js) into the plugin.

To be clear, you want me to do both LightTable/Javascript#9 and LightTable/LightTable#1507 in a plugin?

joshuafcole:
Hmm, fair enough.
I'm not able to get this working with either chromium (even with `--disable-web-security`) or FF though, and I'm using Ubuntu, so I don't have access to Safari.

Can somebody with a mac please confirm this? The code looks fine, but I'd rather see it in action before merging.

This is the (expected) cors error I get with chromium and FF:
```
XMLHttpRequest cannot load http://localhost:60280/socket.io/lighttable/ws.js. Origin file:// is not allowed by Access-Control-Allow-Origin.: file:///home/josh/tmp/test.html
```

rigdern:
Does @cldwalker or @one-more-minute have a Mac to test on?

You should be able to test on a Mac as described in https://github.com/LightTable/LightTable/pull/1507#issuecomment-67093210.

rigdern:
@joshuafcole Do you need help with the testing? I can verify that the new functionality works on an additional Windows machine and you can verify that there's no regression in the existing functionality on Ubuntu.

PR #1513, opened 2014-06-07
Allow to specify filetype depending on file name, not only by extension

vladimir-vg:
Now you can specify `:filenames` option in your behaviors file to assign file type depending on file name, not only by extension.

For example here I ask to highlight `Gemfile` and other files as ruby code:
```clojure
:files [(:lt.objs.files/file-types [{:name "Ruby"
                                     :filenames ["Gemfile" "Rakefile" "config.ru" "Capfile" "Guardfile"]
                                     :exts [:rake]
                                     :mime "text/x-ruby"
                                     :tags [:editor.ruby]}])]
```

Please look at code, I can't name myself as a Clojure programmer; so maybe pull-requested code is not so idiomatic.

### DIFF
--- a/src/lt/objs/files.cljs
+++ b/src/lt/objs/files.cljs
@@ -18,11 +18,15 @@
 
 (defn typelist->index [cur types]
   (let [full (map (juxt :name identity) types)
-        ext (for [cur types
+        exts (for [cur types
                   ext (:exts cur)]
-              [ext (:name cur)])]
+              [ext (:name cur)])
+        filenames (for [cur types
+                        filename (:filenames cur)]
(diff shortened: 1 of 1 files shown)


cldwalker:
@vladimir-vg Thanks for the PR. As we realized in #1668, :exts already works for filenames so there isn't a reason to add this. For example, [this Ruby plugin uses it for ruby files](https://github.com/existentialmutt/lt-ruby/blob/7a2eb8091ac1617727a39b379e70d1a515a38dae/ruby.behaviors#L20). I'm closing but happy to reopen if we find a motivating use case for adding :filenames

PR #1606, opened 2014-08-26
Set width of inputs / files in sidebar

seancaffery:
In a project with a deeply nested folder structure, renaming the file results in a very small input. This ensures the input is at least as wide as the filename being edited.
There is also a visual change to that maintains the highlight width consistent even when the sidebar needs to scroll.


cldwalker:
I verified this is still effective against master and there isn't a clear way to do this with skinning since we're dynamically extending widths
@seancaffery Better late than never. Thanks!

PR #1690, opened 2014-10-26
DRY: comment/uncomment/toggle-comment

Bost:
BTW in 6cf6d31b37dfaf52c4f0a48f97c32114b60d729a @brabadu modified a bunch of normal functions to arity-overloaded functions in order to get rid of some warnings by passing nil as a default parameter. I think this way we just stash away the warnings without properly resolving them. IMO it would be better to leave the warnings displayed until someone gets irritated enough (as I got) and fixes the warnings properly, just as I did in this patch.


cldwalker:
Thanks! Better late than never

PR #1743, opened 2014-11-21
fix warn 'rm: deploy/light: No such file or directory' in oxs_deps.sh

smallfish:


cldwalker:
Thanks for the interest but this script won't be used in our next release. We're redoing scripts as part of switching to atom-shell

Issue #1728, opened 2014-11-20
Investigate using the atom-shell instead of node-webkit


ibdknox:
We've had a lot of issues with nw lately and it might be worth looking into using the wrapper that atom uses for chromium.

PR #1756, opened 2014-11-30
Atom Shell

one-more-minute:
Figured I'd turn Chris's work on #1728 into a PR so we can all discuss what needs doing and make changes. Feel free to edit this with new tasks.

- [x] Sort out inconsistencies with quitting (Atom, at least on OS X, distinguishes between closing windows and quitting).
- [x] Porting CLI
      - [x] drag/drop opening
      - [x] open files
      - [x] open from file manager during and after startup
- [x] Save and restore window position
- [x] Previous workspace isn't loading on startup
- [x] Windows Testing
- [x] Linux Testing
- [x] Packaging Build Step
- [x] Port nw package.json options to atom-shell. References: [nw options](https://github.com/rogerwang/node-webkit/wiki/Manifest-format) and [browser-window options](https://github.com/atom/atom-shell/blob/8278d8b51a6ad38a309dc33ec42b5d06ed459465/docs/api/browser-window.md#new-browserwindowoptions)
     - [x] Figure out why icon option doesn't work
- [x] Sort out JS loading (JQuery isn't working for me any more)
- [x] Keybinding issues
  - [x] Esc doesn't work in command bar
- [x] Resolve error loading node modules when `pwd` is not the app directory
- [x] File save dialogs are broken
- [x] ~~Project search leaves working indicator on~~ (ok, my workspace is just really huge)
- [x] Julia plugin
  - [x] Interactive plots broken
- [x] platform/open no worky
- [x] Binary and app updates
- [x] script/build.sh (to replace *_deps.sh)
      - [x] build on osx and functions correctly
      - [x] clone plugins
      - [x] copy over executables e.g. light
      - [x] handle plugins/node/node{,.exe}
      - [x] rename app - may need to workaround [this issue](https://github.com/atom/atom-shell/issues/960) on osx
      - [x] git add and use relevant contents of [zip files](https://github.com/LightTable/LightTable/blob/ee937ce8548ba6243d6f613925f217d6cdc2ec4e/osx_deps.sh#L16)
           - [x] Mac: port plist
      - [x] osx issues: console printing is wonky, ~~icon doesn't show up~~, ~~initial focus is inconsistent~~
      - [x] build on linux - done on 64bit ubuntu 14.04
      - [x] build on windows - done on 32 bit windows 7
- [x] mac menu issues: ~~Select All and Cut don't work.~~ This issue also exists with node-webkit and only happens when vim is enabled.
- [ ] Port proxy support (we don't have a way to test this)
- [x] Evaling into browser tabs
- [x] Windows issues
      - [x] ~~cursor issues (doesn't change when resizing tabs)~~ - not observed in windows 7
      - [x] closing last window doesn't close app
      - [x] opening file from cygwin terminal doesn't work
- [x] Linux issues
      - [x] opening files from commandline doesn't work
      - [x] ~~worker syntax error~~ - not observed in ubuntu 14
- [ ] Updates to LT should be driven by main.js so that they can be guaranteed to only happen once

Install Instructions: 
See doc/developer-install.md for full instructions

For OSX:
```sh
$ script/build.sh
# open from commandline
$ builds/lighttable-0.8.0-mac/light
# OR open executable
$ open -a $PWD/builds/lighttable-0.8.0-mac/LightTable.app
```

For Linux:
```sh
$ script/build.sh
# open from commandline
$ builds/lighttable-0.8.0-linux/light
# OR open executable
$ open -a $PWD/builds/lighttable-0.8.0-linux/LightTable
```

For Windows:
```sh
$ script/build.sh
# open executable
$ builds/lighttable-0.8.0-windows/LightTable.exe
```


cldwalker:
@ibdknox For plist I opted to copy the one from the zip file since we had a complicated definition for CFBundleDocumentTypes

cldwalker:
@ibdknox For auto updating, I plan on using the [tags api](https://api.github.com/repos/LightTable/LightTable/tags) to detect the latest version. For binary detection, I'll add writing the atom-shell version to a version file and have that replace deploy/version-url. As for the above item "Updates to LT should be driven by main.js so that they can be guaranteed to only happen once", could you explain how that guarantees happening once? Converting deploy.cljs to main.js seems like a fair amount of work

gekkoe:
@ibdknox Much nicer with atom. The slowdown had gotten very bad for my Mavericks machine.

That said, to get it to build with build.sh I had to add sudo to the npm install and grunt lines. Otherwise I got permission errors when it would attempt to mess with things in my /opt folder. 

Runs great after that though.

cldwalker:
@ibdknox Using an http proxy, I tried [deploy/proxy?](https://github.com/LightTable/LightTable/commit/eb526db55c5bf0e0fb81b08d0fdb00f2faef3108#diff-a27d831e5730304bc2b99611ccb6b7bbL41) and only saw it work for "http://temp2.kodowa.com.s3.amazonaws.com/playground/releases/" but not "https://d35ac8ww5dfjyg.cloudfront.net/playground/releases/". Also, doesn't return "PROXY ..." for the github api endpoints. Any objection if I delete it or I am using it wrong?

ibdknox:
@cldwalker the problem is that get-proxy needs to be fixed for atom-shell. The function should only return a proxy if you have a proxy setup locally. Is that what you were expecting?

ibdknox:
Oh and re: updating, the thought was that since there's one master coordinating process, it could handle determining if an update is needed and send it off to one of the windows to handle. Right now ensuring that only one window does it is prone to races as it tries to go through localstorage. I guess we don't have to change it in order to get things working again, it's just something that I always knew was wrong and the only fix was some coordinating process (which we now have). Another fix is to leave it all in the windows and do some message passing - that's probably safer than the localstorage hack.

cldwalker:
> The function should only return a proxy if you have a proxy setup locally. Is that what you were expecting?
I've setup SOCKS and http proxies locally and `get-proxy` didn't return "PROXY ..." for the above urls except for http://temp2.kodowa.com so I thought it was unused functionality. Have any instructions for how you run a proxy so I can test this?

> Right now ensuring that only one window does it is prone to races as it tries to go through localstorage.

How about we wrap the [check-version behavior](https://github.com/LightTable/LightTable/blob/eb526db55c5bf0e0fb81b08d0fdb00f2faef3108/src/lt/objs/deploy.cljs#L193-L196) in a first window check to prevent that?

ibdknox:
I've never tried to setup a proxy, so I'm not much help there. But get-proxy as it is right now is just an empty def, right? atom-shell has a new way of checking for proxies that I didn't implement.

> How about we wrap the check-version behavior in a first window check to prevent that?

The first window may no longer be open.

cldwalker:
> I've never tried to setup a proxy, so I'm not much help there. But get-proxy as it is right now is just an empty def, right? atom-shell has a new way of checking for proxies that I didn't implement

Switching to atom-shell's `app.resolveProxy` is easy enough but without a means of verifying it we don't actually know if it works. How about we ask users who use a proxy to try our alpha release as is? If there isn't an issue then no need for get-proxy and if there is then they should be able to explain a setup we can recreate.

> The first window may no longer be open.

Yep. Thought that would be marginally better than a race condition but maybe not. I don't think this is as high priority as what remains but if we could prevent a race condition that would be nice.

script/build.sh
```diff
@@ -0,0 +1,64 @@
...
+
+# Check if npm is installed
+npm --version >/dev/null 2>&1 || { echo >&2 "Please install npm before running this script."; exit 1; }
+
+# Ensure we start in project root
+cd "$(dirname "${BASH_SOURCE[0]}")"; cd ..
+
+# Ensure we have current version of atom-shell
+pushd shell
+  npm install -g grunt-cli
```
brabadu: I just got this error. Grunt manual says it may require sudo access to install `grunt-cli`. 

```
~/projects/lighttable/shell ~/projects/lighttable
npm ERR! Darwin 14.0.0
npm ERR! argv "node" "/usr/local/bin/npm" "install" "-g" "grunt-cli"
npm ERR! node v0.10.33
npm ERR! npm  v2.1.8
npm ERR! path /usr/local/bin/grunt
npm ERR! code EACCES
npm ERR! errno 3

npm ERR! Error: EACCES, unlink '/usr/local/bin/grunt'
npm ERR!  { [Error: EACCES, unlink '/usr/local/bin/grunt'] errno: 3, code: 'EACCES', path: '/usr/local/bin/grunt' }
npm ERR!
npm ERR! Please try running this command again as root/Administrator.
npm ERR! error rolling back Error: EACCES, unlink '/usr/local/bin/grunt'
npm ERR! error rolling back  { [Error: EACCES, unlink '/usr/local/bin/grunt'] errno: 3, code: 'EACCES', path: '/usr/local/bin/grunt' }

npm ERR! Please include the following file with any support request:
npm ERR!     /Users/brabadu/projects/lighttable/shell/npm-debug.log
```

kolya-ay: I don't think that sudo is the cleanest solution. As a workaround grunt [could be executed](https://github.com/gruntjs/grunt/issues/687) without file system pollution (e.g. `npm install grunt-cli ; node_modules/.bin/grunt download-atom-shell`). However I think that installing grunt especially for atom-shell download is a temporary quick & dirty solution..

joshuafcole: I agree with Kolya on this actually, I'd just recommend not installing it globally.

cldwalker: Updated build script so this isn't an issue

cldwalker:
deploy/run.sh has been moved to script/light.sh

Mouvedia:
If you are planning to break most plugins you should maybe choose a bunch of them, fork them, fix them and move them all under a repo that you will continue to maintain.

kolya-ay:
The core as it is still requires much work to dissipate the force to the foreign plugins maintaining. I propose to extend plugin metadata so that plugin's author could specify supported LT versions. Something like [npm's engine](https://docs.npmjs.com/files/package.json#engines), may be..

kenny-evitt:
@one-more-minute it seems like the "Windows issues" and "Linux issues" checklist items should be considered complete too. [I'm having trouble building on Windows myself](https://github.com/LightTable/LightTable/pull/1892#issuecomment-105305951) but otherwise I think we should consider this done.

I think we should punt on proxy support for now.

I'm not sure about the last item – "Updates to LT should be driven by main.js so that they can be guaranteed to only happen once". Given the existing comments for this, I think we should punt on this for now too.

@cldwalker @ibdknox @one-more-minute – if possible, it would be great if builds could be released as an alpha version for the rest of the community to test.

kenny-evitt:
@cldwalker @ibdknox @one-more-minute @joshuafcole @rundis 

There are two items not marked complete:

 - [ ] Port proxy support (we don't have a way to test this)
 - [ ] Updates to LT should be driven by main.js so that they can be guaranteed to only happen once

Can we defer completing them before creating a new release? Can defer them for a beta release?

TheInitializer:
is there a way to port it back to nw? The performance is generally a lot better there

kenny-evitt:
@TheInitializer No need; just use an old version (there's a recent issue with the links). Or fork the code prior to the merge of this the *atom-shell* branch. I'm not interested in maintaining two versions of the app, nor do I suspect are any of the other commiters or contributors. But I can offer tentative, sporadic, and minimal help for anyone that is so interested.

PR #1798, opened 2015-01-03
Update README to note OS/X Yosemite performace issues

timgilbert:
This patch just adds a note in the README about the performance issues in OS/X Yosemite, just to warn users until the atom-shell version is released.


cldwalker:
Thanks for the concern but not looking for this. Like others issues, users can find out about the state of an issue by searching in [github issues](https://github.com/LightTable/LightTable/issues?q=is%3Aissue+is%3Aopen+yosemite) or the [mailing list](https://groups.google.com/forum/#!searchin/light-table-discussion/yosemite). We also don't want to point users to stale instructions. Instructions are being kept up to date elsewhere as was explained in my last email. We'll be cutting a pre-release in the next week for users who can't build the atom-shell branch

Issue #1793, opened 2014-12-28
Autocomplete token leaks to layout


brabadu:
I have a clojure file with compojure routes.

```
(defroutes myroutes
   ...
  (route/not-found "<h1>404</h1>"))
```

When I was changing code in some other place, I typed / and got autocomplete box with tag h1 applied to text. 

![light_table](https://cloud.githubusercontent.com/assets/177366/5565034/9497fa7a-8ee9-11e4-8f7d-34d47d0b5042.png)

PR #1803, opened 2015-01-04
Escape autocomplete hints #1793

brabadu:
It took me couple of evenings to dig into this. May need more changes.

Autocomplete plugin relies heavily on `lt.objs.sidebar.command/filter-list` for rendering, sorting etc. So I had to find a proper place to insert escaping. 

This commit escapes autocomplete tokens, so html tags aren't interpreted as tags anymore.  BUT now keys for tokens with escapable symbols can't by used for autocompletion.

Example, in my file I have string "http://fb.me/react-0.12.2.js". Before this change I could type `fb.me/` and still see this token in autocomplete list. With this PR symbol `/` is interpreted as `&#x2F;` so it doesn't match and this line doesn't show up in autocomplete. 

Same for a string that caused a bug in first place. Now I can't autocomplete string `<h1>404</h1>` by typing < ( plugin reads it as `&lt;`). It will find this line if you start with `h1` or `404` though.

Anyway I don't think autocomplete should produce such options like `<h1>404</h1>`. It should give only tag or only text, not whole string.

### DIFF
--- a/src/lt/plugins/auto_complete.cljs
+++ b/src/lt/plugins/auto_complete.cljs
@@ -9,7 +9,7 @@
             [lt.objs.editor :as editor]
             [lt.objs.context :as ctx]
             [clojure.string :as string]
-            [lt.util.js :refer [wait]]
+            [lt.util.js :refer [wait escape]]
             [lt.util.dom :as dom])
   (:require-macros [lt.macros :refer [behavior defui background]]))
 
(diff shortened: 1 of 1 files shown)


joshuafcole:
Yeah, autocomplete giving `<h1>404</h1>` doesn't feel very nice, but I don't think making it impossible to complete across escapable symbols is going to be an acceptable solution. We should filter the displayed text only.

cldwalker:
Agreed we should only filter the displayed text. (filter-list) takes a :transform option to do this. The transform fn takes 4 args but it's the third arg that we care about here. The :transform option applied to this issue looks like `:transform #(escape %3)`. Strangely the autocomplete doesn't [render correctly](https://www.dropbox.com/s/0ky9uchebe055uo/Screenshot%202015-01-28%2021.56.55.png?dl=0). It may be an issue with escape or how [dom/html renders an autocomplete item](https://github.com/LightTable/LightTable/blob/21fcb13c45a474fc011c7137eccde4e2662cec9b/src/lt/objs/sidebar/command.cljs#L116-L120). Anyone who wants to dig into this more please do.

cldwalker:
Closing as the suggested approach hasn't been attempted. Happy to reopen if there's interest

PR #1811, opened 2015-01-08
Makes inline result address more specific to fix #1213 and...

joshuafcole:
... LightTable/Clojure#32


cldwalker:
Thanks for looking into this. If the :widgets format has changed shouldn't we update other references as well? I'm seeing a handful of the old format `[line type]` in eval.cljs, doc.cljs and watches.cljs. Shouldn't we cut a LT release now that the plugin depends on master changes?

PR #1813, opened 2015-01-11
Change osx_deps.sh/windows_deps.sh to curl over https instead of http

kkroening:
`osx_deps.sh` and `windows_deps.sh` are using insecure http transport to download dependencies. This PR changes the scripts to use https.

Note that `linux_deps.sh` is already using https.


cldwalker:
Thanks for the interest but these scripts won't be used in our next release. We're redoing scripts as part of switching to atom-shell

PR #1837, opened 2015-01-27
Add SCons file types to load Python behavior

sawcordwell:


sawcordwell:
Not sure if it would be best to add a new line or add the SCons filenames to the Python entry, I thought this seemed nicest. Also, not sure about the order with the captial letter.

cldwalker:
Thanks for the pull but this should be done against the Python plugin
@ibdknox This is introducing a library-specific filetype for lang which we haven't done before. Do we want to start doing this?

ibdknox:
I feel like the best thing to do is to have plugins for libraries. It seems
like a slippery slope trying to put library specific stuff into the
language plugins directly.

cldwalker:
@sawcordwell Thanks for the PR but I agree with Chris that having language plugins track all possible library extensions could be a never-ending task. I recommend putting this behavior in your user.behaviors or if you'd like to share with others publish it as a plugin

sawcordwell:
fair enough, I did put it in my user.behaviors file. Thanks for the info

PR #1840, opened 2015-01-29
Fix css injection-eval filename escaping.

ohAitch:
Found/fixed during hack night.
CSS files being evaluated are injected as a style tag with an id derived from the file name; the derivation was not properly escaping any characters besides `.`, such as ` `.


joshuafcole:
:+1: Thanks for the contribution! It was fun working with you, and you're a pro at finding bugs.

Issue #1810, opened 2015-01-08
add command : toggle word wrap


dcz-switcher:
I use LT in production for web dev (front end) and it's great, I think there is not much to add to have the next editor for devs

A feature which may be cool is the ability to toggle word wrap from the command panel

joshuafcole:
This seems reasonable to me. I'll tag it as a beginner issue so anyone who wants to contribute can tackle it. It should be as simple as dynamically adding/removing the `:lt.objs.editor/wrap` behavior from the active editor.

mjgpy3:
Unless I'm misunderstanding something, it looks like this already exists here:
https://github.com/LightTable/LightTable/blob/master/src/lt/objs/editor/pool.cljs#L579-L585

joshuafcole:
So it does! Thanks for pointing that out mjgpy3, I'll mark this issue closed.

snufkon:
I didn't know `:editor.force.wrap` command to read this issue. After I read this issue, I tried to use this command and I find following unexpected behavior.

- When I open command pane, line wrapping state change to "Wrapping on".

I think this is a bug and if so, I will send a pull request.

PR #1844, opened 2015-01-31
fix a bug of :editor.force.wrap command

snufkon:
I wrote a comment about this bug(maybe bug...) in the end of this issue(#1810).

I did two things.

- fix a bug.
- add message notifying wrap on and off to indicate which one is done.

### DIFF
--- a/src/lt/objs/editor/pool.cljs
+++ b/src/lt/objs/editor/pool.cljs
@@ -581,8 +581,14 @@
               :exec (fn []
                       (when-let [ed (last-active)]
                         (if (editor/option ed "lineWrapping")
-                          (object/add-tags ed [:editor.force.unwrap])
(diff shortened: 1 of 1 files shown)


cldwalker:
I don't see the wrapping state change when the command pane is opened. Perhaps opening the command pane makes it seem the state changed because it changes the width of an already wrapped tab. Adding a notification message is helpful.

As for explicitly removing tags, I do observe that both :editor.force.wrap and :editor.force.nowrap stick around in editor atoms which seems undesirable. But this doesn't seem to be effecting correct behavior, perhaps to due to exclusive behaviors these tags map to. @ibdknox Should we be removing tags for this case?

ibdknox:
yeah I believe we should be removing them.

cldwalker:
Thanks!

PR #1846, opened 2015-01-31
added new bindings

thewhitetulip:
1. toggle comment with ctrl-/
2. toggle workspace with ctrl-\
3. add browser tab with ctrl-b
4. add tabset with ctrl-t
5. remove tabset with shift-ctrl-t
6. using alt-[0-9] to switch between tabs rather than ctrl-[0-9]
7. toggle console by alt-c


cldwalker:
Hi. Changes to LightTable's default keybindings are considered when they are recognized as standard keybindings used by other apps across linux, windows and osx. Most of these seem to belong in your user.keymap. For each keybinding you'd like considered please explain how it's a default as was done in #1434.

thewhitetulip:
1. toggle comment with ctrl-/ 
This is done in almost all the code editors including but not limited to Eclipse, atom, sql developer, I think mysqlworkbench I am not sure though, it is a standard way to toggle comments, it makes life easier really.

2. toggle workspace with ctrl-\
This isn't a standard toggle workspace keybinding of any editor that i am aware of till date except Atom, the thing is normally in old code editors you didn't toggle the workspaces but here you might need to, while coding django I create two tabsets, one with the code and another with the browser to see the o/p inline

3. add browser tab with ctrl-b
This is a new binding and it truly is lighttable specific

4. add tabset with ctrl-t
5. remove tabset with shift-ctrl-t these two are new bindings as well

6. using alt-[0-9] to switch between tabs rather than ctrl-[0-9]
This I am pretty sure all apps do, I am not sure in the Windows/Mac, but in the Linux world all the tabs can be navigated with alt-[0-9] (in firefox, chrome, terminal, geany, I think in eclipse etc)

7. toggle console by alt-c
this is a totally new binding, but a useful at that

thewhitetulip:
1. alt-up alt-down moves the line/selected lines up or down
 a very standard and handy feature of most text editors
2. alt-d duplicates the line/line selection
this is standard across editors Eclipse and Geany at least

cldwalker:
keybinding 1 is the only keybinding here that is used across multiple apps. 3,4 and 7 would conflict with vim and emacs keybindings that users use through plugins. keybinding 6 breaks standard keybindings for mac and windows users. As mentioned above, most of those keybindings are more appropriate for user.keymap

Can you flatten this pull to one commit for just the comment toggle keybinding and use "pmeta-/" instead of "ctrl-/"? Using pmeta ensures command key for osx and the ctrl key for other linux and windows. If I don't hear back soon I can do this as well.

thewhitetulip:
I have no idea how to flatten a pull request, could you please let me know how to do it?

cldwalker:
Just make a new pull request for the comment toggle keybinding in one commit and I can close this one

thewhitetulip:
https://github.com/LightTable/LightTable/pull/1868

PR #1850, opened 2015-02-04
Restored note of OSX differences in hotkeys

gekkoe:
While it's not as useful a place for this as an up-front note on the README like it used to be (admittedly that document was getting a bit long), it's still important to document for OSX users that their hotkeys are not consistent with the ones talked about in the documentation, and let them know how to find out which they should be using.


cldwalker:
Thanks for the PR but I just opted to fix the references

PR #1864, opened 2015-02-24
Update license with up-to-date year

fay-jai:
Only made 1 change:
1) Update LICENSE.md to include the latest year in the form of a date range.

No other changes were made.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,7 +1,7 @@
 The MIT License (MIT)
 =====================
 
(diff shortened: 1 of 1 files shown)


cldwalker:
Thanks for the pull. Unfortunately changing licenses isn't a one-project edit and should be made across all our projects. I'd like to avoid updating all licenses every year and just leave it as is. If you're looking for ways to contribute, see [here](https://github.com/LightTable/LightTable/blob/master/CONTRIBUTING.md#contributing)

Issue #1796, opened 2015-01-01
Define file types for codemirror modes


cldwalker:
We should provide syntax highlighting for all the [codemirror modes](https://github.com/LightTable/LightTable/tree/master/deploy/core/node_modules/codemirror/mode) we load at startup time. This would prevent issues like #1795. In order to do this, we need to add file-typed declarations for each mode. Steps to do this:

1. For each mode [in this directory](https://github.com/LightTable/LightTable/tree/master/deploy/core/node_modules/codemirror/mode), check to see if there is [a file type definition](https://github.com/LightTable/LightTable/blob/b5a71aac3b06d29b5fe93625b37347e4757467b2/deploy/settings/default/default.behaviors#L270-L316). The following modes should _not_ have a definition  because they are defined in plugins: "clojure" "css" "htmlembedded" "htmlmixed" "javascript" "python"
2. If there isn't, add one in the correct spot (they're alphabetically sorted). A definition consists of the following keys: 
    
* :exts - List of file extension keywords for that mode e.g. [:edn] for edn mode
* :mime - mime type string for a mode. This can be found at the bottom of each mode definition. For example, [apl is "text/apl"](https://github.com/LightTable/LightTable/blob/b5a71aac3b06d29b5fe93625b37347e4757467b2/deploy/core/node_modules/codemirror/mode/apl/apl.js#L173)
* :tags - List of keywords in the format :editor.lower-case-name e.g. :editor.apl.
* :name - Capitalized string name for a mode

To verify that syntax highlighting works for a mode, try an example from [Codemirror's mode examples](http://codemirror.net/mode/) and use the `Set current editor syntax` command to change the syntax on your current tab. A pull request on this is most welcome!

justintaft:
Quick and dirty way to solve the issue:
- wget http://codemirror.net/mode/  & follow all the mode links
- for each page, sed  & grep  the necessary parts , print parts out to a file formatted as clojure data structure

PR #1866, opened 2015-02-27
Added file types for codemirror modes

alhimik45:
As described in the #1796 issue, I added all definitions


cldwalker:
Thanks for knocking this out!

PR #1899, opened 2015-05-02
Electron name change: update for Linux references

miraleung:
Addressing Linux-specific references in pull request #1892.

PR #1910, opened 2015-05-28
Switch from atom-shell to electron (new name of a-s)

Armaklan:
Name of atom-shell has changed.

Actually, build don't work without this modification (unable to get atom-shell dependencies).


Armaklan:
Double of PR 1889
https://github.com/LightTable/LightTable/pull/1899

Sorry...

Issue #1903, opened 2015-05-07
LT 0.7.2 - Workspace doesn't display directory if name contains the string ".git"


Dadie:
Example directory names: 
- foo.git
- foo.git.bar
- foo.github.io
- .git.bar
- .git

Interestingly, if you create those directories via the workspace contextmenu they actually show up but wont appear on later starts. If created externally they never show up.

reproduces on x86_64 Arch Linux.

rundis:
LT is default setup to ignore certain file patterns. You can override this in your user behaviours if you so wish.

[:app :lt.objs.files/file.ignore-pattern "(^\\..*)|\\.class$|target/|svn|cvs|\\.git|\\.pyc|~|\\.swp|\\.jar|.DS_Store|\\.nrepl-port"]

Dadie:
So the LT default setup is so that every file or directory that contains the substring ".git" is ignored? Is this really on purpose? I see no reason to ignore every file or directory that contains the substring ".git", so what was the reason for this decision? 

I only see two possibilities. The first is that it is on purpose, so why is it on purpose? And the second, that it isn't on purpose and was meant to just ignore the meta-file directory ".git". In that later case it is obviously a bug that any directory or file that contains the substring ".git" is also ignored.

I can change my default setting, no problem with that. But I just feel uncomfortable with this solution, as I'm not sure whether to see this as a "bug" or as "as intended".

kenny-evitt:
@Dadie It's probably a bug. Pull requests welcome!

Issue #1739, opened 2014-11-20
Update ClojureScript dependency


igorhub:
User plugins can't use many modern libraries since Light Table depends on ClojureScript 0.0-2138, which is 10 month old.

cldwalker:
This issue has been superseded by #1973

Issue #1647, opened 2014-10-03
Adding a test suite to Light Table


cldwalker:
EDIT: Adding a test library like clojurescript.test to Light Table isn't an issue. Trying to write acceptance tests on any importance namespace that depends on node-webkit/electron is an issue.

To address issues in #1646, I attempted to add a test suite to Light Table but ran into issues.

First, I tried adding clojurescript.test on [this branch](https://github.com/cldwalker/LightTable/tree/add-cljs.test-tests). However, when running tests with `lein cljsbuild test`, the compiled js errors on references to global, caused by references to js/global.

Example output:

``` sh
$ lein cljsbuild test
... Tons of warnings

Successfully compiled "target/cljs-test.js" in 29.125813 seconds.
Running ClojureScript test: test
ReferenceError: Can't find variable: global

  target/cljs-test.js:17876

Testing lt.behaviors-test 

ERROR in (lt.behaviors-test/failing-test) (:)
expected: (seq (clojure.core/deref lt.object/behaviors))
  actual: 
TypeError: 'undefined' is not an object (evaluating 'lt.object.behaviors')
    at target/cljs-test.js:14646
    at target/cljs-test.js:14359
    at target/cljs-test.js:14385
    at target/cljs-test.js:14420
    at target/cljs-test.js:14241
    at target/cljs-test.js:14422
    at target/cljs-test.js:14241
    at target/cljs-test.js:14429
    at target/cljs-test.js:14442
    at target/cljs-test.js:14459
    at target/cljs-test.js:14470
    at target/cljs-test.js:14502
    at target/cljs-test.js:14527
    at target/cljs-test.js:6186
    at target/cljs-test.js:6234
    at target/cljs-test.js:14535
    at target/cljs-test.js:14543
    at phantomjs://webpage.evaluate():2
    at phantomjs://webpage.evaluate():8
    at phantomjs://webpage.evaluate():8

Ran 2 tests containing 2 assertions.
Testing complete: 0 failures, 1 errors.

Ran 2 tests containing 2 assertions.
Testing complete: 0 failures, 1 errors.
Subprocess failed
```

Then I attempted to simply add tests that just call `assert` on [this branch](https://github.com/cldwalker/LightTable/tree/add-assert-tests). Executing the resulting js file with `node target/test.js` resulted in `ReferenceError: HTMLCollection is not defined` which appears in dom.cljs.

[...]

PR #1921, opened 2015-06-17
#1903 Fixed default.behaviors file.ignore-pattern

Dadie:
For more information see #1903

### DIFF
--- a/deploy/settings/default/default.behaviors
+++ b/deploy/settings/default/default.behaviors
@@ -52,7 +52,7 @@
  [:app :lt.objs.workspace/init-workspace-cache-dir]
  [:app :lt.objs.metrics/init-metrics]
  [:app :lt.objs.keyboard/chord-timeout 1000]
- [:app :lt.objs.files/file.ignore-pattern "(^\\..*)|\\.class$|target/|svn|cvs|\\.git|\\.pyc|~|\\.swp|\\.jar|.DS_Store|\\.nrepl-port"]
+ [:app :lt.objs.files/file.ignore-pattern "^\\.|\\.class$|^target/|svn|cvs|\\.pyc$|~$|\\.swp$|\\.jar$"]
  [:app :lt.objs.style/provide-skin "dark" "core/css/skins/new-dark.css"]
  [:app :lt.objs.style/provide-skin "light" "core/css/skins/new-dark.css"]
  [:app :lt.objs.style/provide-skin "new-dark" "core/css/skins/new-dark.css"]


kenny-evitt:
Thanks for the PR!

But yikes – this seems a lot more ambitious than what your issue covers.

You need to escape `\b` as `\\b` – the former will be interpreted as a backspace character.

I don't understand some of the changes:

 - `(^\\..*)` to `^\\.(.*)` – honestly the original version seems to cover anything beginning with a period so the subsequent more-specific patterns seem superfluous.

Maybe you could split these changes into separate commits?

It'd be really nice to have some tests for this, i.e. a readable list of example file names that will be ignored.

As-is, let me review this for a bit and test that this is a sensible change.

@rundis – would you take a look at this too?

Dadie:
First of all thanks for the quick reply. 
And thanks for having patient with me.

About the \b as \\b : 

* Missed that, sorry.

About ```(^\\..*)``` to ```^\\.(.*)``` : 

* I though the brackets where kind of "useless" as they didn't add anything to help to 
visualize what the RegEx is doing. The later one separates the first period from the second which makes th regEx (IMO) quite easier to read and to understand.

About tests:

* And how should this test look like? I could create a list, but how exactly should I test that list? (As I don't see any kind of test code in the project)

About more commits:

* I'll make for each term a single commit.

Dadie:
I rewrote most of the code, split the commit in many small commit (each addressing a 'bug')

kenny-evitt:
@Dadie thanks for the contribution!

The capturing group (i.e. the parentheses) in the first sub-pattern – `(^\\..*)` is extraneous and can be removed.

The patterns `\\.git`, `.DS_Store`, and `\\.nrepl-port` patterns can be removed.

The remaining 'dot' patterns – `\\.class$`, `\\.pyc$`, `\\.swp$`, and `\\.jar$` – should be retained.

`~` should be probably be `~$` too.

You're right that there are no tests. But for now it'd be nice if we just test manually, i.e. create a defined set of files and directories and confirm that they're visible or not in Light Table in the workspace.

Dadie:
* Removed \\.git, \\.DS_Store, and \\.nrepl-port 370e869
* ~ to ~$ 4baf569
* Removed extraneous parentheses in file.ignore-pattern (^\\..*) 501d0d4

Tests for pattern ```^\\..*```
* http://regexr.com/3b7tl

Test for pattern ```\\.class$```
 * http://regexr.com/3b7to

Test for pattern ```\\.pyc$```
* http://regexr.com/3b7tr

Test for pattern ```~$```
* http://regexr.com/3b7tu

Test for pattern ```\\.swp$```
* http://regexr.com/3b7u1

Test for pattern ```\\.jar$```
* http://regexr.com/3b7u4

Test for pattern ```^target/```
* http://regexr.com/3b7ud

kenny-evitt:
The first pattern can be simplified to `^\\.`.

Besides that, I think the changes are good. I want another committer to sign-off before merging. And contrary to what I told you earlier, I'd now rather this be squashed into a single commit.

@rundis What do you think about these changes?

kenny-evitt:
@Dadie note that you're going to need to change one of your local behaviors files for now as we don't know when we're going to be able to make the next release.

Dadie:
@kenny-evitt 

The first pattern can be simplified to ^\\. f07c586

Should I squash it now into a single commit or should I better wait for the other committer?

Also: my local behaviors is already loaded with the updated version :) But thanks for reminding me.

kenny-evitt:
@Dadie You can go ahead and squash now; thanks.

Dadie:
@kenny-evitt I squashed the commit

rundis:
Looks good to me (but then again my regex foo is not up to scratch either).

kenny-evitt:
@Dadie There's no real hurry to merge this as the core focus is to get the Atom/Electron version tested and released.

But I think this is a good opportunity to explore how to best add testing. I created [a branch in my LT fork](https://github.com/kenny-evitt/LightTable/tree/atom-shell-with-a-few-tests) for playing around with this. I started trying to get [*clojurescript.test*](https://github.com/cemerick/clojurescript.test) working.

In the source file *src/lt/objs/files.cljs*, I added a function `ignore?` that extracts a simple expression used in several places for excluding files based on the file-ignore pattern. I was thinking of adding a few tests with a very simple file-ignore – e.g. `#"a"` – to demonstrate that any file for which *any* portion matches the pattern will be ignored. The remaining tests will cover the default pattern.

Sorry in advance for the stupid mistakes in the test code I committed on that branch.

kenny-evitt:
Bad news on adding tests – see #1647 for the first documented attempt:

 - [Adding a test suite to Light Table · Issue #1647 · LightTable/LightTable](https://github.com/LightTable/LightTable/issues/1647)

I haven't gotten clojurescript.test to work for me, and I'm pinned to version *0.3.1* as newer versions require newer versions of ClojureScript than Light Table can use and updating ClojureScript is currently blocked by other changes that should be made first – see #1739:

 - [Update ClojureScript dependency · Issue #1739 · LightTable/LightTable](https://github.com/LightTable/LightTable/issues/1739)

[Or, maybe we will need "to run the tests in LT itself".](https://github.com/LightTable/LightTable/issues/1647#issuecomment-58197749)

Any help is very much welcome.

sbauer322:
This looks solid to me @kenny-evitt! I went through the regexr links and added each file/folder to a workspace in LT and compared the results - they were as desired.

I think we should merge this given that automated testing has stalled out for the time being... unless this PR was being held up from something else.

Issue #1789, opened 2014-12-25
"incorrect header check" error installing plugins


jasonzoladz:
Here's the error message.  
--------------------------------------------------------------------------------------------
Error: incorrect header check at Zlib._binding.onerror (zlib.js:295:17)
--------------------------------------------------------------------------------------------

(Is this a git issue?)

cldwalker:
Closing since I can't reproduce this without more information

mikebelanger:
Same error when I try and install Gist ([https://github.com/harsha-mudi/lt-gist/issues/new]).  I used fresh install settings by removing everything in ~/Library/Application Support/Lighttable.  I'm on OS X Yosamite 10.10.2, using Lighttable build 0.7.2.

So start with a new install of Lighttable (from command-line) and simply try and install the Gist plugin from the plugin manager.  I get this error:
```
Error: incorrect header check
    at Zlib._binding.onerror (zlib.js:295:17)
```

Interestingly, there's a similar problem installing the haskell plugin too:

https://github.com/jetaggart/light-haskell/issues/52

kolya-ay:
[The same]((https://github.com/otijhuis/paredit-plus/issues/5) with paredit-plus with a new install.

mbernat:
Still happens on 0.7.2.

troydm:
+1 happens when I try to install Zenburn theme plugin on 0.7.2 running on Ubuntu 14.04 64-bit Linux

kenny-evitt:
@jasonzoladz @mikebelanger @kolya-ay @mbernat @troydm can you consistently reproduce this error? Do you get it when installing *any* plugin or is it specific to specific plugins?

troydm:
@kenny-evitt I haven't tried installing any other plugin but I constantly get this error when I try to install this Zenburn theme plugin so it's reproducible

kenny-evitt:
@troydm thanks

kolya-ay:
Confirm I can't still install paredit-plus through plugin manager. However I can install e.g. Subpar plugin. It seems that [the culprit](https://github.com/LightTable/LightTable/blob/master/src/lt/objs/deploy.cljs#L24) is the [obsoleted npm-dependency](https://github.com/LightTable/LightTable/tree/master/deploy/core/node_modules/tar). I use LT form the `atom-shell` branch.

kenny-evitt:
@kolya-ay thanks

[...]

PR #1937, opened 2015-07-02
Better handling of #1789 - Plugin download errors now handled.

rundis:
Fixed so that downloading of plugins verifies that download succeeds before trying to untar. #1789 indicated that the error was in the gunzip/untar phase, but the problem was really that the download failed and the error response was streamed to file for then to be unzipped etc.. no wonder that failed.

In addition another fix was made. If people add a suffix of .git to the :source attribute in their plugin.edn or plugin.json, the download of plugin tar also fails. Must have worked previously, but apparently not any longer. Anyways any .git to the source path is now stripped off.


kenny-evitt:
@rundis thanks!

PR #1946, opened 2015-07-29
Update developer-install.md

justiniac:
Insufficient rights for npm to build.

### DIFF
--- a/doc/developer-install.md
+++ b/doc/developer-install.md
@@ -45,6 +45,14 @@ $ cd LightTable
 $ script/build.sh
 ```
 
+If `npm` complains about unlocking unlocked stuff, run the following commands and start again.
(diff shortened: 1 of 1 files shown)


kenny-evitt:
@justiniac The info you added doesn't seem to necessarily be good general advice; see [this answer](http://stackoverflow.com/a/24404451/173497) to the SO question to which you linked.

justiniac:
Thanks for the tip (would have spared me some headaches ^^) Is someone taking care of documenting this here?

kenny-evitt:
You took care of documenting this. Luckily, our users are pretty sophisticated and likely to find an existing issue that covers the problem they're having. If anything, add the specific error you observed to make it easier for others to find this pull request. Or create and close an issue for it instead.

If someone else runs into this, I'll reconsider merging your change.

PR #1954, opened 2015-08-24
Add 'package' shell script for creating release package.

kenny-evitt:
Currently the script only supports creating the release package for Windows.


cldwalker:
Closing since we ended up agreeing on adding windows zip support to script/build-app.sh

Issue #1961, opened 2015-08-30
Client processes not cleaned up after killing(/quitting) LT on electron


rundis:
If I connect to say Clojure projects in Light Table and leave them open before quitting LT, the processes (lein-light + light runner) spawned for the projects will be left running after the main LT process is gone.

Might be an issue with other clients too.

rundis:
Seems the trigger :closing is never raised.
I also found that the prompt for closing when you have unsaved files never is invoked either.
Seems to be a close event missing somewhere to tie all this together...

cldwalker:
Thanks for digging in. It seems I caused it with https://github.com/LightTable/LightTable/commit/9658450e3b2a9d61c706d42056656347708e9bab. I remember making this change and asking Chris if he had any objection but never heard back. When I revert this commit I do see Clojure processes killed :) Would be great if we could bring back a solution that doesn't require `windowClosing`. $IPC_DEBUG may come in handy here

cldwalker:
Fixed with #1964

PR #1964, opened 2015-09-01
#1961 Ensure that clean-up behaviors are raised on close and prompts …

rundis:


cldwalker:
@rundis Nice to see this fixed without a global, race-conditioning var! I noticed the [destroy documentation](https://github.com/atom/electron/blob/master/docs/api/browser-window.md#windestroy) only recommends it for crashed scenarios. atom appears to [only use destroy in crashes](https://github.com/atom/atom/blob/8f1daad4a9b606a617206a170f604c828c94ab35/src/browser/atom-window.coffee#L131-L142). Related, I see that we have an [unbeforeunload](https://github.com/LightTable/LightTable/blob/master/src/lt/objs/app.cljs#L230) defined which appears to have no effect (regardless of whether we destroy or not). We should delete it

We could bring back Mike's [extra closeWindow channel](https://github.com/LightTable/LightTable/commit/dffbc93328488f4539506768bc26e595825a5d50) and use a window boolean e.g. window.isLTDoneClosing instead. We can also stay with destroy but if so we should make the [force? true branch](https://github.com/rundis/LightTable/blob/window-close-handling/src/lt/objs/app.cljs#L37-L40) not call .close after .destroy. While it doesn't happen b/c the window is destroyed, it's confusing to read. I leave the decision to you :)

As an aside, I noticed that if a tab is modified and a user accidentally closes the window LT doesn't prevent the user from closing which results in lost work for anyone not using Reminisce. Something to fix down the road I imagine.

rundis:
@cldwalker changed and rebased. I opted for keeping the destroy function as opposed to (re-)introducing yet another global boolean

LightTable's "event" model and electron doesn't quite fit for these lifecycle cases, so there's a bit of event ping-pong and working around the mismatch. I'm sure there is a better solution waiting to reveal itself. I couldn't find any more docs on why destroy shouldn't be used and haven't been able to detect any obvious negative side-effects yet (other than the niggly feeling that electron might have had a really good reason for putting that disclaimer into the api doc).

btw this solution should also prompt users if they try to close with unsaved work (works on my machine ...)

If you are happy with the latest changes, perhaps it would be a good idea if @kenny-evitt gave it a quick go on windows machine before merging.

cldwalker:
> I opted for keeping the destroy function as opposed to (re-)introducing yet another global boolean

Just to clarify, the alternative to destroy didn't involve global vars. `window.isLTDoneClosing` would be a boolean on a BrowserWindow object. Setting booleans on objects is the same way atom deals with closing except that it subclasses BrowserWindow.

> LightTable's "event" model and electron doesn't quite fit for these lifecycle cases, so there's a bit of event ping-pong and working around the mismatch.

I don't think LT's model is in conflict with window lifecycles. When we want to use OOP and access a window object in main.js [we can](https://github.com/LightTable/LightTable/blob/a0802cf07b1029f42f9ba05b381616a429bc7dc6/deploy/core/main.js#L20) and when we want to be functional in cljs we can [as long as we pass the window number](https://github.com/LightTable/LightTable/blob/1d9dd0cc3154048834aedd554a056acb0a2df127/src/lt/objs/dev.cljs#L14)

> btw this solution should also prompt users if they try to close with unsaved work (works on my machine ...)

Hmm. When I edit a file and then type Ctrl-Q, invoke close window from the command bar or select Close window from the menu, LT closes without prompting for the unsaved file. What are you modifying and how are you closing LT?

> If you are happy with the latest changes, perhaps it would be a good idea if @kenny-evitt gave it a quick go on windows machine before merging.

I'm happy with it :) @kenny-evitt what say you?

kenny-evitt:
@rundis @cldwalker It works for me – merge away!

rundis:
@cldwalker opening a file, edit it, press x on the window or quit from the menu. So for you it closes the connections, but not giving you the prompt ? That's a worry. 
https://github.com/LightTable/LightTable/blob/1d9dd0cc3154048834aedd554a056acb0a2df127/src/lt/objs/app.cljs#L100 Should trigger file save prompting right ? 


@kenny-evitt did you get a prompt if you tried quitting light table with a modified file open ?

kenny-evitt:
@rundis Yeah, I got a prompt for every means of closing or quitting LT (and closing the tab directly too) – clicking the red X button, <kbd>Alt</kbd>+<kbd>F4</kbd>, and *File* (menu) > *Close window*.

cldwalker:
@rundis My bad on a false alarm. This was a behavior caused by the Reminisce plugin. I've filed https://github.com/one-more-minute/Reminisce/issues/8. Please merge this! :tada: :tangerine:

cldwalker:
On Ubuntu 14, I can consistently produce the following error when I close LT with 3 or more tabs. I think this is related to the destroy we're using. I don't think this is an issue yet but if enough users are confused/concerned about this we may need to revisit
```
The error was 'BadWindow (invalid Window parameter)'.
  (Details: serial 1610 error_code 3 request_code 40 minor_code 0)
  (Note to programmers: normally, X errors are reported asynchronously;
   that is, you will receive the error a while after causing it.
   To debug your program, run it with the --sync command line
   option to change this behavior. You can then get a meaningful
   backtrace from your debugger if you break on the gdk_x_error() function.)
[9380:1023/000543:ERROR:channel.cc(300)] RawChannel read error (connection broken)
```

rundis:
Maybe #2017 also is related to this too ?
I've seen once or twice some exceptions when closing a browser tab that I've gotten an exception that might have been related too. Unfortunately I didn't keep the stacktrace and try to reproduce.

kenny-evitt:
I was able to reproduce the same error as @cldwalker also on Ubuntu 14.04.

PR #1969, opened 2015-09-09
Optionally zip release files after app build (for Windows).

kenny-evitt:
Zip release files after app build for Windows.

@cldwalker @rundis I finally got around to this. The existing archive files referenced by the "download" links on the home page all contain a *LightTable* directory. The zip file produced by these changes will contain a directory named like *lighttable-0.8.0-windows* (which is fine with me).

I'll merge this if no one objects in a day or two.

### DIFF
--- a/script/build-app.sh
+++ b/script/build-app.sh
@@ -52,9 +52,10 @@ BUILDS=builds
 RELEASE="$NAME-$VERSION-$OS"
 RELEASE_DIR="$BUILDS/$RELEASE"
 RELEASE_TARBALL="$BUILDS/${RELEASE}.tar.gz"
+RELEASE_ZIP="${RELEASE}.zip"
 RELEASE_RSRC="$RELEASE_DIR/$RESOURCES"
 
-rm -rf $RELEASE_DIR $RELEASE_TARBALL
+rm -rf $RELEASE_DIR $RELEASE_TARBALL $RELEASE_ZIP
 
 #----------------------------------------------------------------------
 # Copy Electron installation and app directory into output location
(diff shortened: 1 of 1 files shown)


script/build-app.sh
```diff
@@ -110,4 +111,10 @@ if [ "$1" == "--tarball" ]; then
   tar -zcvf $RELEASE_TARBALL $RELEASE_DIR
 fi
 
+if [ "$1" == "--zip" ]; then
+  pushd builds
+  "/cygdrive/c/Program Files/7-Zip/7z.exe" a $RELEASE_ZIP "$RELEASE/*"
```
cldwalker: Since this is a cygwin specific path, could you make a note that this branch is windows-specific?

script/build-app.sh
```diff
@@ -52,9 +52,10 @@ BUILDS=builds
 RELEASE="$NAME-$VERSION-$OS"
 RELEASE_DIR="$BUILDS/$RELEASE"
 RELEASE_TARBALL="$BUILDS/${RELEASE}.tar.gz"
+RELEASE_ZIP="${RELEASE}.zip"
```
cldwalker: Can we use $BUILDS here and get rid of the pushd and popd? When any paths change it's helpful to only have to change one variable instead of multiple locations in the script

script/build-app.sh
kenny-evitt: Yes; is a comment sufficient?

cldwalker: Yep

script/build-app.sh
kenny-evitt: I can definitely use `pushd $BUILDS` but if I change this line to `RELEASE_ZIP="$BUILDS/${RELEASE}.zip"` then the zip file is created with the files nested under e.g. *builds\lighttable-0.8.0-windows* instead of *lighttable-0.8.0-windows* and the former is ugly (tho possibly consistent with the tarball).

cldwalker: Gotcha. No need to change

cldwalker:
@kenny-evitt Looks good! :+1: :tada:

PR #1976, opened 2015-09-14
Destroy devtools-client along with parent browser - close #1975

### LINKED ISSUE #1975: Closing internal browser not cleaning up properly ?
If I;
1. Open a html file
2. Select command ```Editor: Open current file in browser```
3. Close the browser
4. Select command: ```Console: Clear console```

I get the following exception:
```WebSocket is already in CLOSING or CLOSED state.
bootstrap.js [35090]	lt.objs.clients.devtools.send_STAR_```

I tried reverting back to this [commit](https://github.com/LightTable/LightTable/commit/e97278f174e1b83fffb2f455800530f08a5cdcc1) but the error still persists.

cldwalker:
@rundis I was able to replicate the issue by opening and closing a browser tab or with your instructions. Since the clear-console command looks for [all devtools-client objects](https://github.com/LightTable/LightTable/blob/1d9dd0cc3154048834aedd554a056acb0a2df127/src/lt/objs/console.cljs#L255) and tries to communicate with them, it makes sense to destroy the browser's devtools object when the browser object was destroyed.

### DIFF
--- a/src/lt/objs/browser.cljs
+++ b/src/lt/objs/browser.cljs
@@ -139,6 +139,7 @@
                   :triggers #{:close}
                   :reaction (fn [this]
                               (object/raise this :inactive)
+                              (object/destroy! (:devtools-client @this))
                               (object/destroy! this)))
 
 (behavior ::rem-client


rundis:
Yup. Seems to do the trick! Tx.

PR #1989, opened 2015-09-20
Fix workspace behaviors

cldwalker:
While trying to use the workspace behaviors command, I noticed it was buggy and out of date. Buggy because opening it kicked off incorrect dialogs about file being deleted or prompting to save to a file. Neither of these are correct as workspace behaviors are saved to the current-workspace object.
After this was fixed, I noticed that workspace.behaviors doesn't support the flat format. Given that all docs point to the new format and that autocompleting behaviors only works for the flat format, I fixed this. While fixing this I verified there are only two places where we need to convert/parse :ws-behaviors - when saving them and when loading them when LT starts up

@rundis @kenny-evitt Could I get one off of you to sign off?

### DIFF
--- a/src/lt/objs/settings.cljs
+++ b/src/lt/objs/settings.cljs
@@ -106,14 +106,17 @@
          (sort-by first)
          (vec))))
 
-(defn parse-file [file]
-  (let [behs (-> (files/open-sync file)
-                 :content
-                 (safe-read file))]
-    (cond
+(defn parse-behaviors [behs file]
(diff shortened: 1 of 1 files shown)


rundis:
Smoketested ok: 
- opened workspace behaviors through command bar
- added a behavior override for indentation of javascript (with flat format)
- verified that behavior was indeed loaded
- modified behavior again and checked that new behavior was picked up

Issue #1944, opened 2015-07-19
Misleading error message when uninstalling plugin


ShayMatasaro:
using 0.8 from master.
when uninstalling light refactor i get the following msg:

Some plugin dependencies are missing.

We found that the following plugin dependencies are missing:
clj-light-refactor 0.1.6
Would you like us to install them?

cldwalker:
HI @ShayMatasaro. I've verified this issue. It is misleading message caused by behaviors probably kicking off out of order. Note the uninstall still works. Regardless, it is something we'll need to fix

cldwalker:
This is fixed on master

PR #1990, opened 2015-09-21
Fix installing a plugin doesn't load new behavior- close #1970

### LINKED ISSUE #1970: Commands/behaviors aren't immediately available after installing a plugin through the plugin manager
To reproduce:
* Install a plugin e.g. [Whitespace plugin](https://github.com/LightTable/Whitespace)
* Expect to see a command from the installed plugin in the command bar e.g. Command `Editor: Toggle visible whitespace"` but command not found

If you execute the command `App: Reload behaviors`, then the command is found. This is loosely related to #1042. For an example of a behavior not being immediately loaded see LightTable/Vim#61

cldwalker:
@kenny-evitt @rundis This fixes the long-standing bug with an installed plugin not showing new behaviors, commands and keymap :) @kenny-evitt You should be able to install Vim as you did on the issue you filed and just have it work. Fix explained in commit

### DIFF
--- a/src/lt/objs/plugins.cljs
+++ b/src/lt/objs/plugins.cljs
@@ -492,8 +492,8 @@
            (dom/prevent e)
            (dom/stop-propagation e)
            (discover-deps plugin (fn []
-                                   (cmd/exec! :behaviors.reload)
-                                   (object/raise manager :refresh!)))))
+                                   (object/raise manager :refresh!)
+                                   (cmd/exec! :behaviors.reload)))))
 
 (defui install-button [plugin]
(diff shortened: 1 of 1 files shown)


rundis:
Tried a couple of plugins and their commands was shown in the command bar and executed ok as far as I could tell (without actually being very familiar with any of them).

(uninstalling is a bit annoying with the dialog popping up, but thats a separate issue I guess).

cldwalker:
@rundis Thanks for QAing. Yep, #1944 is tracking the uninstall issue

PR #1992, opened 2015-09-22
Fix uninstall message

### LINKED ISSUE #1944: Misleading error message when uninstalling plugin

cldwalker:
@kenny-evitt @rundis This fixes #1944 which I had thought was fixed back in 7b3ebf1a768d835b195c59b620d0cbd512de8349. I also cleaned up `available-plugins` as having the options dynamic in one place and hardcoded in another was making this confusing to read. By breaking them into more smaller usable chunks less option passing is needed. When QAing, I tried edge cases like deleting a plugin directory and then seeing if the manager would prompt me to install it

### DIFF
--- a/src/lt/objs/plugins.cljs
+++ b/src/lt/objs/plugins.cljs
@@ -446,7 +446,8 @@
 (defn uninstall [plugin]
   (when (:dir plugin)
     (files/delete! (:dir plugin))
-    (object/raise manager :refresh!)))
+    ;; :ignore-missing b/c uninstalled shows up missing
+    (object/raise manager :refresh! :ignore-missing true)))
 
 ;;*********************************************************
 ;; Manager ui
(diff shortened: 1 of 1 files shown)


src/lt/objs/plugins.cljs
```diff
@@ -411,42 +411,47 @@
...
-                            [:span "Would you like us to install them?"]]
-                     :buttons [{:label "Cancel"}
-                               {:label "Install all"
-                                :action (fn []
-                                          (install-missing missing?))}]}))
-    final))
+                     (filterv identity))]
+    (-> (reduce (fn [final p]
+                  (if-let [cur (get final (:name p))]
+                    ;;check if it's newer
```
rundis: this comment seems reduntant :)

rundis:
Nice ! Looks good to me.

(On a separate note, at some point it would be interesting to discuss what would be required to remove a plugins associated commands from the command bar when uninstalling a plugin. Well then there is keymaps and behaviors files with potential references as well... might be to much to chew over)

cldwalker:
Thanks for the review. Yeah, I was also thinking of uninstall doing that. Feel free to make an enhancement issue for that

Issue #2004, opened 2015-10-05
Inline results not cleared when evaling same statement again


rundis:
Steps to reproduce

- Open a editor for clojurescript (say a plugin cljs file)
- connect to lt-ut
Enter and evaluate:
```clojure 
(defn hello []
  (println "hello"))
```
- then change change hello to hello world and voila to results, and so on.

<img width="498" alt="screen shot 2015-10-05 at 22 45 30" src="https://cloud.githubusercontent.com/assets/399197/10293075/ddf6af6c-6bb2-11e5-9d2f-91ceead83022.png">

Happens with both clojure and clojurescript

kenny-evitt:
I was able to reproduce this on Windows just now and I'm pretty sure I saw the same on Linux, both with the latest source.

I wasn't sure this was fixed before.

rundis:
I can also reproduce with javascript (:
![screen shot 2015-10-06 at 07 15 52](https://cloud.githubusercontent.com/assets/399197/10300686/20387fb2-6bfa-11e5-8726-0dbdd960a945.png)

rundis:
Think this commit https://github.com/LightTable/LightTable/commit/2b0e9086927f5870966b9937efbf1c0e6ae47d95#diff-7bc0013fd8e35583a4681d44ccf83a91R313 is the culprit breaking inline results for "normal eval".

I can't see how this could possible work using ch, as soon as you change something on the line or newlines the lookup will be off ? I tried reverting the effects of this commit (or rather just removing the ch specifier for lookup and creation and it seems to work nicely.

But that I guess would mess up instarepl (:

cldwalker:
@rundis I had [similar concerns](https://github.com/LightTable/LightTable/pull/1811#issuecomment-69330596) but never heard back. After the alpha release, it may be worth seeing if changing to `[line type ch]` and making the ordering change in Clojure is doable

rundis:
Fixed on master

PR #2008, opened 2015-10-06
Fix results stacking of inline-results for normal eval

rundis:
Ensure inline results are cleared when evaling lines where col for result is offset from previous eval result Addressing - #2004

Seems to still work as expected with instarepl and watches as well.

### DIFF
--- a/src/lt/objs/eval.cljs
+++ b/src/lt/objs/eval.cljs
@@ -305,14 +305,13 @@
                       (let [ed (:ed @this)
                             type (or (:type opts) :inline)
                             line (ed/line-handle ed (:line loc))
-                            ch (:ch loc)
                             res-obj (object/create ::inline-result {:ed this
                                                                     :class (name type)
                                                                     :opts opts
                                                                     :result res
                                                                     :loc loc
                                                                     :line line})]
-                        (when-let [prev (get (@this :widgets) [line ch type])]
+                        (when-let [prev (get (@this :widgets) [line type])]
                           (when (:open @prev)
                             (object/merge! res-obj {:open true}))
                           (object/raise prev :clear!))
@@ -320,7 +319,7 @@
                           (doseq [widget (map #(get (@this :widgets) [(ed/line-handle ed %) type]) (range (:start-line loc) (:line loc)))
                                   :when widget]
                             (object/raise widget :clear!)))
-                        (object/update! this [:widgets] assoc [line ch type] res-obj))))
+                        (object/update! this [:widgets] assoc [line type] res-obj))))
 
 ;;****************************************************
 ;; underline result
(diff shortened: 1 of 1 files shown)


kenny-evitt:
This seems good. It worked for Clojure, ClojureScript, and JS files, and a Clojure InstaREPL.

kenny-evitt:
I say merge away.

But do we need to, or should we, re-QA again?

cldwalker:
I don't see how it's working for certain instarepl cases if https://github.com/LightTable/Clojure/commit/301729825c4e0117c59d719606142cfd603834c3 is still in Clojure. I'd rather not do this until we've understood what was fixed in instarepl and keep the instarepl fix in while fixing our eval. If we want to get this in before alpha, we'll need to at least QA related behaviors afterwards

rundis:
The Inline result behavior is used by lot's of plugins (clojure, javascript, instabuster, clojure-light-refactor, python, haskell, etc etc). All of them I potentially suffering from the instarepl fix (or rather the dependent fix in eval.cljs in LT core).

From what I can figure out, it doesn't seem the clojure instarepl is actually using the behavior from eval, it rolls it's own. So I'm guessing(/hoping perhaps) that any potentially negative ramifications of this pull would be limited to bordeline cases where you switch from normal repl to instarepl and back, potentially with watches and what not. Allthough to me it seems a full clean of widgets happens when you do this switch anyway.

I 'd love to come up with a definitive convincing story, but I guess I'm not quite there yet.

I did test quite a bit both with and without the instarepl (and watches) for this fix, but of course no guarantees can be made until I actually fully understand the instarepl impl.

To summarize my ramblings:
- Fixing a generic behavior used by potentially lots of plugins should trump potentially breaking some edge cases in the clojure instarepl
- The fix only fixes an annoyance (simple workaround to clean inline results first), not a vital fix by any means so it shouldn't hold back an alpha-release unless we're talking hours or less in difference
- I'd be inclined to shoot from the hip and include it, but then again I'm biased :-)

cldwalker:
@rundis Agreed keeping generic eval is more important than a one-off fix for instarepl. Any reason this PR doesn't just revert 2b0e9086927f5870966b9937efbf1c0e6ae47d95?

rundis:
That would have been to easy :-)

rundis:
maybe this line should stay ? https://github.com/LightTable/LightTable/pull/2008/files#diff-7bc0013fd8e35583a4681d44ccf83a91L485

cldwalker:
I looked into this more and understand what happened. LightTable/Clojure@3017298 is the correct fix for instarepl stacking. Since instarepl/update-res uses inline objects it was thought that eval/inline-results, an inline behavior, also needed to be updated. However, this is false. If you search for the associated :editor.result trigger it is only found in clojure.cljs which is used by clj + cljs project, _not_ instarepl. I observed this to be true with print debugging.

> maybe this line should stay ? https://github.com/LightTable/LightTable/pull/2008/files#diff-7bc0013fd8e35583a4681d44ccf83a91L485

Given this commit was wrong in other ways, I'm ok with also reverting this until we observe an issue with 'Clear inline results'. It seems to work fine in clj, cljs and instarepl

@rundis Thanks for ringing the alarm and please merge! :+1:

rundis:
Fixed #2004

PR #2042, opened 2015-11-16
Introduce block comment command  and make toggle comment block aware

rundis:


kenny-evitt:
This seems pretty low-risk to me.

I'll try to actually test it sometime too :smile:

cldwalker:
@rundis Nice find! Block comments by default is nice for the existing toggle comment command.
Is there a specific case you have in mind for :block-comment-selection or is it more to be consistent with :comment-selection and :uncomment-selection? For example in Javascript, the only ability it would give a user is the ability to `/* code */` a single line instead of `// code`. Are there users who want this? If so, wouldn't they also expect to be able to have that preference with :toggle-comment-selection? 
 This is less a reflection of your PR and more thinking out loud if these extra comment commands (:comment-selection and :uncomment-selection) which don't compose are useful to users

rundis:
I pretty much only use toggle-comment.
I added block-comment-selection for consistency  

If we disregard the toggle command, there was no support for block-commenting (only uncommenting blocks) before this pull. 

I can think of an edge-case where block commenting a single line is useful: In elm doc strings require block comments, regardless of whether it's one or more lines ! 

There might be a case for making the behavior of toggle-comments configurable if a user always wishes to use line comments regardless of 1 or more lines. (but then again there is a workaround to use comment/uncomment commands ...)

cldwalker:
> I can think of an edge-case where block commenting a single line is useful: In elm doc strings require block comments, regardless of whether it's one or more lines !

Good to know

> There might be a case for making the behavior of toggle-comments configurable if a user always wishes to use line comments regardless of 1 or more lines. (but then again there is a workaround to use comment/uncomment commands ...)

Yeah. I guess we'll see if users want this configurable. I'd forgotten but we could use #1690 to do this later

Looks good to merge! :+1:

PR #2046, opened 2015-11-18
Fix version sorting for building plugin cache data. Fixes #2043

### LINKED ISSUE #2043: Plugin Manager - False "Last version"
There is a sort of problem in the version numbers. Sometimes the proposed version is not the latest version.

Exemple with TernJs.

Last version in LightTable : 0.1.9
Real last version : 0.1.10

Metadata repository is correct.

kenny-evitt:
Previously `build-cache` was attempting to sort plugin versions but [`sort-by`](https://clojuredocs.org/clojure.core/sort-by) was being called with the version comparator being passed as the sort key function. The version comparator was also inverting the standard return values for a comparator – see [`compare`](https://clojuredocs.org/clojure.core/compare) for an example.

### DIFF
--- a/src/lt/objs/plugins.cljs
+++ b/src/lt/objs/plugins.cljs
@@ -197,8 +197,8 @@
 (defn version-sort [a b]
   (cond
    (= a b) 0
-   (deploy/is-newer? a b) 1
-   :else -1))
+   (deploy/is-newer? a b) -1
+   :else 1))
 
 (defn build-cache [sha]
   (let [items (filter files/dir? (files/full-path-ls metadata-dir))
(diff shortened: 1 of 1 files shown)


kenny-evitt:
@cldwalker @rundis This seems pretty minor but it would be nice to include in the next release.

src/lt/objs/plugins.cljs
```diff
@@ -207,7 +207,7 @@
                           :let [versions (->> (files/full-path-ls plugin)
                                               (filter files/dir?)
                                               (map plugin-info)
-                                              (sort-by #(version-sort % %2))
+                                              (sort-by #(:version %) version-sort)
```
cldwalker: `#(:version %)` can just be :version

cldwalker:
@kenny-evitt Nice sleuthing and thanks for digging into this! I'm surprised sort was ever working without :version. Feel free to merge after that minor suggestion

src/lt/objs/plugins.cljs
kenny-evitt: Good point

kenny-evitt:
@cldwalker This seems to have worked because the directory enumeration must have been effectively ordering the directories by name.

PR #2051, opened 2015-11-30
Improve new user experience

cldwalker:
Over the past couple weeks, I've had some designers (non backend people) at my coworking space trying out LightTable. The feedback was pretty consistent in that user.keymap and user.behaviors explanations were confusing (they don't know what diffs and objects are). To address this I revamped the user.keymap and user.behaviors explanations to be more practical and created a separate document for describing configuration in detail. I also added menu items for functionality users were looking for but not finding since they weren't familiar with the command pane. Even when they found the command pane, these new users preferred to access them through the menu. Given configuration/settings and the plugin manager are integral parts of using LightTable, I think adding them to the menu will help new users.

### DIFF
--- a/deploy/core/User/user.behaviors
+++ b/deploy/core/User/user.behaviors
@@ -16,7 +16,7 @@
 ;; * Optional arguments are arguments to pass to a behavior. If a behavior has arguments they should
 ;;   pop up in an info box after the behavior has been autocompleted.
 ;;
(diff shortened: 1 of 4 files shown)


cldwalker:
@rundis @kenny-evitt Will merge this end of day tomorrow unless there are any concerns

rundis:
Looks good to me.

cldwalker:
Thanks for looking it over!

kenny-evitt:
Nice!

PR #2052, opened 2015-12-01
Upgrade to Electron 0.34.5

cldwalker:
Doing the upgrade because our auto-updating notification doesn't work from 0.x.x-alpha to 0.x.x. Also we're doing a full QA for 0.8.0, so we may as well pull in upstream bug fixes with an upgrade.
Held off on upgrading to 0.35.0 since there are ipc changes to handle. Also worth noting that https://github.com/atom/electron/releases/tag/v0.33.4 moved devtools-* events. Our code still works fine for now but something we'll eventually want to change

I ran through about 1/2 the QA checklist on OSX and found no new issues. @rundis @kenny-evitt Could one of you try some of the QA on linux or windows? We can always revert this if it fails our full QA

### DIFF
--- a/deploy/core/version.json
+++ b/deploy/core/version.json
@@ -1,1 +1,1 @@
-{"version":"0.8.0-alpha","electron":"0.31.1"}
+{"version":"0.8.0-alpha","electron":"0.34.5"}
--- a/deploy/electron/Gruntfile.js
+++ b/deploy/electron/Gruntfile.js
@@ -4,7 +4,7 @@ module.exports = function(grunt) {
   grunt.initConfig({
     pkg: grunt.file.readJSON('package.json'),
     "download-electron": {
-      version: "0.31.1",
+      version: "0.34.5",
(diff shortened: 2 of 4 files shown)


kenny-evitt:
@cldwalker I'll QA Windows in the next few days.

cldwalker:
Cool. Doesn't have to be too comprehensive since we'll be doing that for the release but at least a sanity check that common operations work.
Also, we should switch to using node.js > 4.0 for development after this PR since electron is using node again (now that the two have merged)

rundis:
Ok I went through most of the checklist and everything seemed fine on Ubuntu. 
I did have some small hickups related to browser stuff, (something about a channel closed or already closing.. darn that memory, should have saved the errors...).

kenny-evitt:
I tested this on my (poor, under-powered) Windows 7 box just now, cursorily, and it looks good.

cldwalker:
Thanks for the review!

PR #1955, opened 2015-08-25
Add docstrings per #1715

cldwalker:
Finally got around to this. Had to do code reading for some of this. Hopefully this will make LT more accessible and encourage us to document more :)
@kenny-evitt @rundis  Feedback welcome on anything that's unclear

PR #2059, opened 2015-12-06
Add api docs

cldwalker:
This adds API docs for LightTable building off of https://github.com/LightTable/LightTable/commit/f94bf3af0d382424c69161d5b01680d0ea657a38 so we can encourage plugin authors and more users to read and understand the code base. I've included a script and documentation for publishing these docs. [Here is what the current docs look like](http://lighttable.github.io/LightTable/api/lt.macros.html). The styling isn't amazing but I'm ok with for a first pass. I did have to mark some fns private and added a few more docstring for public fns. I limited docs to a few namespaces as covered in #1955. I'm open to adding more down the road.

There weren't too many choices when choosing between libraries that generate documentation for cljs projects: I only found codox and marginalia. I chose codox because only it supported links to fns e.g. [object/add-tags](http://lighttable.github.io/LightTable/api/lt.object.html#var-add-tags) and I think it's important to encourage conversation around docs.

@rundis @kenny-evitt Would like to get this merged before our release

### DIFF
--- a/src/lt/macros.clj
+++ b/src/lt/macros.clj
@@ -36,7 +36,7 @@
          (lt.util.dom/on e# ev# func#))
        e#)))
 
-(defmacro timed [ev & body]
+(defmacro ^:private timed [ev & body]
   `(let [start# (lighttable.util.js/now)
          res# (do ~@body)]
      (lighttable.components.logger/log ~ev (- (lighttable.util.js/now) start#))
@@ -47,30 +47,30 @@
     [(first body) (rest body)]
     [[] body]))
 
-(defmacro on [name & body]
+(defmacro ^:private on [name & body]
   `(lighttable.command/on ~name (fn ~@body)))
 
-(defmacro in [ctx & body]
+(defmacro ^:private in [ctx & body]
   (let [[params body] (->params body)]
     `(assoc ~ctx :in (fn ~params ~@body))))
 
-(defmacro out [ctx & body]
+(defmacro ^:private out [ctx & body]
   (let [[params body] (->params body)]
     `(assoc ~ctx :out (fn ~params ~@body))))
 
-(defmacro defcontext [name & body]
+(defmacro ^:private defcontext [name & body]
   `(let [ctx# {:name ~name}]
      (lighttable.context/add-context!
        (-> ctx#
            ~@body))))
 
-(defmacro extract [elem kvs & body]
+(defmacro ^:private extract [elem kvs & body]
   (let [defs (vec (apply concat (for [[k v] (partition 2 kvs)]
                                   `[~k (lt.util.dom/$ ~v ~elem)])))]
     `(let ~defs
        ~@body)))
 
-(defmacro foreach [xs & body]
+(defmacro ^:private foreach [xs & body]
   `(let [xs# ~(second xs)
          len# (.-length xs#)]
      (loop [left# 0]
@@ -79,7 +79,7 @@
            ~@body
            (recur (inc left#)))))))
 
-(defmacro with-time [& body]
+(defmacro ^:private with-time [& body]
   (let [start (gensym "start")
         body (walk/postwalk-replace {'time (list '- '(.getTime (js/Date.)) start)} body)]
   `(let [~start (.getTime (js/Date.))]
@@ -98,7 +98,7 @@
         (.unshift args# (.-obj msg#))
         (.apply ~func nil args#)))))
 
-(defmacro aloop [[var arr] & body]
+(defmacro ^:private aloop [[var arr] & body]
   `(let [arr# ~arr]
      (loop [~var 0]
        (when (< ~var (.-length arr#))
--- a/src/lt/object.cljs
+++ b/src/lt/object.cljs
@@ -12,27 +12,27 @@
   (:require-macros [lt.macros :refer [behavior with-time aloop]]))
 
 ;; HEART of BOT Architecture!
-(def obj-id
+(def ^:private obj-id
   "Counter to guarantee unique object ids"
   (atom 0))
 
-(def instances
+(def ^:private instances
   "Map of object ids to objects created by object/create"
   (atom (sorted-map)))
 
-(def behaviors
+(def ^:private behaviors
   "Map of behavior names to behaviors created by macros/behavior"
   (atom {}))
 
-(def object-defs
+(def ^:private object-defs
   "Map of object template keys to template maps created by object/object*"
   (atom {}))
 
-(def tags
+(def ^:private tags
   "Map of tags to associated lists of behaviors"
   (atom {}))
 
-(def negated-tags
+(def ^:private negated-tags
   "Map of tags to dissociated lists of behaviors e.g. :-behavior"
   (atom {}))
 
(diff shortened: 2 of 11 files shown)


src/lt/macros.clj
```diff
@@ -47,30 +47,30 @@
     [(first body) (rest body)]
     [[] body]))
 
-(defmacro on [name & body]
+(defmacro ^:private on [name & body]
```
rundis: Unless you really think these should be private / tagged as private shouldn't you use the :no-doc true meta tag https://github.com/weavejester/codox#metadata-options?

cldwalker: I am aware of :no-doc. My intent is to keep as much private as possible until we know we want them to be public

rundis:
I found [codeina](https://github.com/funcool/codeina) to provide a lot nicer output than codox. Changing from codox to codeina shouldn't be hard they have very overlapping syntax/usage interface.
I've used it with pretty good success on my only cljs lib for general consumption [rewrite-cljs](http://rundis.github.io/rewrite-cljs/).

All though cljs doesn't honor any encapsulation with regards to private, I don't think the doc generation should be the deciding factor for what is considered private or not. Heck as of now the whole of the cljs code in LT could be considered it's public interface. Who knows what plugin authors might have started using. Ah anyway that's a entire discussion on it's own, 

I suggest to use the codox meta data tag :no-doc true, once that is fixed I'd say go for merge.

Btw I think this is a big step in the right direction of making LT more approachable for developers considering hacking away at their editor.

cldwalker:
@rundis codeina looks nice! Do you want to try that in a PR after the release? Some things to about if we switch from codox to codeina: One advantage of codox is that weavejester has a longer track record of maintaining projects than funcool. One disadvantage of codeina is that they forked codox instead of working with them. To me that indicates a maintainer who gets excited in the short term and doesn't maintain their projects in the long term.

src/lt/macros.clj
rundis: Ok that makes sense. We probably should try to narrow down the public api quite a lot to make it maintainable (and a lot easier to make changes in the future without to much fear of breaking stuff).

cldwalker:
@rundis Another option with codeina, is to take their styling and add it codox. weavejester [was very responsive today](https://github.com/weavejester/codox/issues/107) for an issue that I wanted to address around linking to source. Also, I think that is an issue we would have to solve in codeina

rundis:
@cldwalker I guess codox is the closest thing to a defacto std for cljs docs so if we can help out improving that I'm all for that. Got you on the concern about track record and maintenance. Allthough I really hope funcool keeps it up, because we rely heavily on his buddy lib for our sso solution at work :-)

src/lt/object.cljs
```diff
@@ -349,20 +351,20 @@
...
     inst))
 
-(defn refresh! [obj]
+(defn- refresh! [obj]
   (reset! obj (update-listeners obj))
   (raise* obj (trigger->behaviors :object.instant (:tags @obj)) nil)
   (raise obj :object.refresh))
 
-(defn add-behavior! [obj behavior]
+(defn- add-behavior! [obj behavior]
```
rundis: This is used several places outside the namespace to dynamically add behavior (also I use it in one of my plugins to add test related behaviour after an editor has been opened). I guess for symmetry rem-behavior! should also be public.

src/lt/object.cljs
```diff
@@ -40,33 +40,35 @@
...
-(defn ->triggers [behs]
+(defn- ->triggers [behs]
   (let [result (atom (transient {}))]
     (doseq [beh behs
             t (:triggers (->behavior beh))]
       (swap! result assoc! t (conj (or (get @result t) '[]) beh)))
     (persistent! @result)))
 
-(defn specificity-sort
+(defn- specificity-sort
```
rundis: this is used in the keyboard ns (can't believe I remembered that !), but I'm not sure if the function belongs in the object namespace in the first place.

src/lt/object.cljs
```diff
@@ -285,12 +287,12 @@
   [obj & r]
   (swap! obj #(apply update-in % r)))
 
-(defn assoc-in! [obj k v]
+(defn- assoc-in! [obj k v]
```
rundis: I just used recently used this. There is a update! function (that is basically update-in, which would have been a better name btw), so assoc-in!, or assoc! would make sense for me to have at ones disposal.

src/lt/objs/editor.cljs
```diff
@@ -125,23 +125,23 @@
   [e]
   (. (->cm-ed e) (getValue)))
 
-(defn ->token [e pos]
+(defn- ->token [e pos]
```
rundis: I'm pretty sure this is used quite a bit (ref clojure plugin, and I have used it alot)  and needs to be public. Or perpaps there is a better way to get hold of the (cm) token at a given pos ?

src/lt/objs/editor.cljs
```diff
@@ -125,23 +125,23 @@
...
   (js->clj (.getTokenAt (->cm-ed e) (clj->js pos)) :keywordize-keys true))
 
-(defn ->token-js [e pos]
+(defn- ->token-js [e pos]
   (.getTokenAt (->cm-ed e) (clj->js pos)))
 
-(defn ->token-type [e pos]
+(defn- ->token-type [e pos]
```
rundis: I need this for 3 of my plugins !

src/lt/object.cljs
cldwalker: Agree its placement looked weird. Seems we should keep it private until we're sure of its purpose

src/lt/objs/editor.cljs
cldwalker: Not that I know of. Will change

src/lt/objs/editor.cljs
cldwalker: :) Will change

src/lt/object.cljs
cldwalker: Agree update-in is the better name. One day we should deprecate update!. Will change assoc-in!

src/lt/object.cljs
cldwalker: Will do

cldwalker:
@rundis Made suggested changes. Anything else? I've made #2060, #2061 and #2062 which I think captures followup todos to this PR

cldwalker:
@rundis I didn't make an issue for making more fns public for the 7 namespaces. Feel free to make one if there are more you'd like to consider

kenny-evitt:
All of this looks good.

I concur with @cldwalker that it would probably be best to work with Codox instead of Codeina.

As for Marginalia, I'm confused what you meant that it doesn't 'supports links to fns'. AFAIK, the documentation is interleaved with the code so that's not an issue unless you were referring to, e.g. linking to function `bar` from the doc-string for function `foo`.

cldwalker:
> As for Marginalia, I'm confused what you meant that it doesn't 'supports links to fns'. AFAIK, the documentation is interleaved with the code so that's not an issue unless you were referring to, e.g. linking to function bar from the doc-string for function foo

I meant the ability to have an external url to any var/fn e.g. http://lighttable.github.io/LightTable/api/lt.objs.files.html#var-copy. I don't think marginalia has that but perhaps I missed it

rundis:
Looks good !

PR #2068, opened 2015-12-08
Add changelog for 0.8.0

cldwalker:
I took what we had from the alpha release changelog and added a few more bullet points since then.
@kenny-evitt @rundis Can I get a thumbs up?

### DIFF
--- a/deploy/core/changelog.md
+++ b/deploy/core/changelog.md
@@ -1,5 +1,46 @@
 #Changes
 
+## 0.8.0
+
+* CHANGED: We have switched to Electron from NW.js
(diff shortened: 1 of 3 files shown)


cldwalker:
Also noticed that we forgot to bump the documentation url which is needed for source links to point to correct tag on github

kenny-evitt:
@cldwalker :+1:

I added an explicit mention of the Codox `:source-uri` in the version bumping checklist
item.

@salmazov Lots of users were having problems with the NW version. See #1728 for some corroboration.

PR #2071, opened 2015-12-11
Core team should be getting notifications

cldwalker:
@kenny-evitt @rundis Nothing too surprising here. Will merge tomorrow


kenny-evitt:
Useful :smiley:

Issue #1973, opened 2015-09-11
Update ClojureScript


cldwalker:
Updating ClojureScript is important because it's preventing us from using cljs.test (#1647) and more importantly from evaling in modern clj/cljs projects (LightTable/Clojure#52). Here is a running list of known todos needed to update cljs:

* Remove uses of `lt.util.cljs/clj->js` and `lt.util.cljs/js->clj` and rely on the fns of the same name that come with clojurescript
* Update `lein cljsbuild` and fix all compilation warnings. Known warnings to fix include missing requires and prefix JSON e.g. js/JSON
* Recompile cljsDeps.js [like this](https://github.com/BenjaminVanRyseghem/LightTable/commit/e2fffcc9d7e58ddaf48fa58cfaeb3f263136956f)
* May need to use fetch's master for https://github.com/LightTable/fetch/pull/17
* Anything else?

Note that upgrading cljs is being attempted in #1952 and was last attempted by Chris in 8e73f59891c45f73a1b985fed69795d1061a8ecf. For this upgrade to be considered successful, the branch must pass our [QA checklist](https://github.com/LightTable/LightTable/wiki/QA-Checklist)

lks128:
@cldwalker, I would like to participate on solving this issue as it feels like a mayor blocker for such important thing as automated testing.

It looks like long time there was no activity... Are there any blockers?

Regarding the QA checklist it would be nice to automate it, I'm referring: https://github.com/electron/electron/blob/master/docs/tutorial/using-selenium-and-webdriver.md

kenny-evitt:
@lks128 There's a checklist of items at the top of the issue. Also see the referenced issues and PRs for other details.

cldwalker:
@lks128 Great! Attempting to fix any of the Fix items would be appreciated. As outlined in the first items, steps to setup an upgrade environment: upgrade cljs, lein-cljsbuild and `lein cljsbuild once cljsdeps`. We won't be able to merge this into master until the QA checklist works but if any progress is made on the checklist, we could save any work to a separate branch

sbauer322:
>  Make sure extend-types that occur on core classes are updated - see cljs.cljs

@cldwalker or @kenny-evitt, could you clarify what this means, or how I would go about making sure the extend-types are properly updated?

kenny-evitt:
@sbauer322 I don't know what that means. @rundis – thoughts?

sbauer322:
[...]

PR #1952, opened 2015-08-18
Update cljs

BenjaminVanRyseghem:
In order to try another approach for #1949 , I bumped all the dependencies to the latest.

I fixed all the warnings and errors, but the editor is still quite not loading.
If someone wants to have a look at it, it open this PR while I continue working on it :smile:

PR #2072, opened 2015-12-11
Prelude to clojurescript upgrade

cldwalker:
Unlike previous attempts, this PR only address upgrade todos for #1973 that are compatible with the current version of ClojureScript. By doing this upfront work, we make the upgrade branch more manageable and closer within reach :smile: . This PR does the following:

* Fixes all clojurescript warnings that future versions of lein-cljsbuild will pick up. These are warnings that we shouldn't be causing anyway so these are uncontroversial.
* This introduces a second build target for generating cljsDeps.js as done by @BenjaminVanRyseghem in #1952. Having a second build target can introduce unreliability in `lein cljsbuild once`. I'll double check to see if I need to update script/build.sh in some way. This build target only needs to be run when we upgrade cljs.
* Remove our dependence on `js->clj` and `clj->js`. This is necessary for an upgrade as we can't continue to depend on home-rolled versions of such a core fn. This change has some small risk. I"ll run through the QA checklist but I don't expect to find any issues. If you're curious how I determined what plugins use our `js->clj`, it was with [this search](https://github.com/search?l=&q=user%3ALightTable+js-%3Eclj&ref=advsearch&type=Code&utf8=%E2%9C%93).

@kenny-evitt @rundis I'll be using this branch exclusively until we merge. I plan to merge Monday or Tuesday unless you have any concerns

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 /deploy/core/node_modules/lighttable/cljs/
+/deploy/core/node_modules/clojurescript/cljsDeps/
 /target/
 /deploy/plugins/
 .DS_Store
--- a/cljsdeps/core.cljs
+++ b/cljsdeps/core.cljs
@@ -1,0 +1,4 @@
+(ns cljsdeps.core
+  (:require [clojure.string :as string]
+            [clojure.walk :as walk]
+            [cljs.reader :as reader]))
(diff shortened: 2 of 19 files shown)


kenny-evitt:
I'm running this branch now. I'm not doing any work with Clojure or ClojureScript tho.

cldwalker:
I ran through most of the QA checklist and haven't seen any issues. I'll double check cljsbuild reliability and if that's good merge tomorrow

kenny-evitt:
I've been running a build of this branch on a Mac and a Windows computer without issue.

cldwalker:
@kenny-evitt @rundis Fyi. Now that this branch is merged you should run `lein cljsbuild [once|auto] app` when recompiling LT cljs

PR #2086, opened 2015-12-27
fix first-time build to not require bootstrap.js to exist

rgiar:
Otherwise a first-time build seems to succeed but silently fails and hangs at start


cldwalker:
Thanks for noticing this. Fixed

Issue #2003, opened 2015-10-05
Umlaut 'ü' on enabled pair-keymap-diffs


Wolfi:
Related to #620

In version 0.7.2, if pair-keymap-diffs is enabled (`[:app :lt.objs.settings/pair-keymap-diffs]`) the `ü` key inserts `[` instead.

kenny-evitt:
@Wolfi Please try to reproduce this with a build of the latest source version. #620 is tagged as being fixed on *master*.

kenny-evitt:
I see now that the fixed-on-master tag for #620 should mean that this is fixed on 0.7.2 too. But this still should be reproduced on the current source version.

cldwalker:
We were tracking adding support for this in #1807. I'll reiterate what I said there: having pair keybindings work on international keyboards would be great. Pull requests would be appreciated since none of the core team uses (or is too familiar with) international keyboards

TeachMeFly:
Hi everyone. This problem can be solved by using codemirror addon, that allready exist in lighttable package, just add this to user.behaviors to apply this addon:

     ;; Normal breackets autoclose
     [:app :lt.objs.editor/load-addon ["edit/closebrackets.js"]]
     [:editor :lt.objs.editor/set-codemirror-flags {:autoCloseBrackets true}]

And you may disable lighttable autoclose plugin
      [:editor :-lt.objs.settings/pair-keymap-diffs]

This addon also close single quotes(') and add extra new line when you press return inside empty breackets.

kenny-evitt:
@TeachMeFly – interesting! That would be great if this is a fully general solution.

Armaklan:
Work fine on linux and windows with Azerty keyboard (french).
Thx ! :)

TeachMeFly:
@kenny-evitt Why LightTable have it own autoclose-brackets code instead using Codemirror addon in the first place?

kenny-evitt:
@TeachMeFly Answering that would require some archaeology of the code history. I'm just guessing, but it could have been because that CodeMirror addon didn't exist when the LT code was added.

kenny-evitt:
@TeachMeFly Mind submitting a PR to remove the LT code/plugin and add your behavior setting to the default behavior file?

kenny-evitt:
@TeachMeFly Nevermind – just got to the email notification for #2093!

Issue #620, opened 2013-06-07
On a german keyboard the 'ü' key is interpreted wrong.


heyLu:
With a german keyboard layout (`setxkbmap de` on Linux) LightTable ignores the `ü` key and inserts `[` instead. Both are on the same key, but differ depending on the layout.

I don't think this is intentional, but I'm not sure. The key on the right of  the `ü` or `[` results in the correct character (`+` and `]` respectively).

I'm using 0.4.11, but I'm not sure whether this is a new problem or I noticed it only now.

ottonascarella:
on japanese keyboard on mac book pro
all the keys are swapped. caps key is interpreted as control and so forth.
this way I would have to have different shortcuts for different layouts, even though I wanna keep same layout.

PR #2093, opened 2016-01-03
#2003 Add behaviors to enable/disable edit/closebrackets.js addon.

TeachMeFly:
Add enable/disable behaviors codemirror edit/closebrackets.js addon, that fix pairs autocomplete for international keyboards.

For more information see #2003

### DIFF
--- a/deploy/settings/default/default.behaviors
+++ b/deploy/settings/default/default.behaviors
@@ -195,6 +195,7 @@
  [:editor :lt.objs.editor/hide-line-numbers]
  [:editor :lt.objs.editor/hide-fold-gutter]
  [:editor :lt.objs.editor/no-wrap]
+ [:editor :lt.objs.editor/disable-brackets-autoclose]
  [:editor :lt.objs.settings/default-keymap-diffs]
  [:editor :lt.objs.settings/user-keymap-diffs]
  [:editor :lt.objs.settings/default-behavior-diffs]
--- a/src/lt/objs/editor.cljs
+++ b/src/lt/objs/editor.cljs
@@ -702,6 +702,7 @@
           :triggers #{:init}
           :reaction (fn [this]
                       (load/js "core/node_modules/codemirror/addon/edit/matchbrackets.js" :sync)
+                      (load/js "core/node_modules/codemirror/addon/edit/closebrackets.js" :sync)
                       (load/js "core/node_modules/codemirror/addon/comment/comment.js" :sync)
                       (load/js "core/node_modules/codemirror/addon/selection/active-line.js" :sync)
                       ;; TODO: use addon/mode/overlay.js
@@ -749,3 +750,19 @@
                         (load/js "core/node_modules/codemirror/addon/display/rulers.js" :sync))
                       (let [rulers (or rulers [{:lineStyle "dashed" :color "#aff" :column 80}])]
                         (set-options this {:rulers (clj->js rulers)}))))
+
+(behavior ::enable-brackets-autoclose
+          :triggers #{:object.instant :lt.object/tags-removed}
+          :desc "Editor: Enable brackets autoclose"
+          :exclusive [::disable-brackets-autoclose]
+          :type :user
+          :reaction (fn [this]
+                      (set-options this {:autoCloseBrackets true})))
+
+(behavior ::disable-brackets-autoclose
+          :triggers #{:object.instant :lt.object/tags-removed}
+          :desc "Editor: Disable brackets autoclose"
+          :exclusive [::enable-brackets-autoclose]
+          :type :user
+          :reaction (fn [this]
+                      (set-options this {:autoCloseBrackets false})))
--- a/src/lt/objs/editor.cljs
+++ b/src/lt/objs/editor.cljs
@@ -756,8 +756,12 @@
           :desc "Editor: Enable brackets autoclose"
           :exclusive [::disable-brackets-autoclose]
           :type :user
-          :reaction (fn [this]
-                      (set-options this {:autoCloseBrackets true})))
+          :params [{:label "map"
+                    :example "{:pairs \"()[]{}''\\\"\\\"\" :explode \"[]{}\"}"}]
+          :reaction (fn [this opts]
+                      (if opts
+                        (set-options this {:autoCloseBrackets (clj->js opts)})
+                        (set-options this {:autoCloseBrackets true}))))
 
 (behavior ::disable-brackets-autoclose
           :triggers #{:object.instant :lt.object/tags-removed}


cldwalker:
Thanks for the pull. Solving pairing for international keybindings would be great. I noticed that the addon closes single quotes (') which is not desired as a default [as explained here](https://github.com/LightTable/LightTable/commit/82fe8cbfc8001c550182b8ae900c022d003e46c5). Should be easy enough to change by passing [a pairs option](https://github.com/LightTable/LightTable/blob/863a7b289eab2a6d2539865bcad565410a15285f/deploy/core/node_modules/codemirror/addon/edit/closebrackets.js#L25) to the addon. However the bigger issue with this pull is that we need to think through these questions to verify closebrackets.js works for more international keyboards and how we will introduce it:

* Does closebrackets.js work for other international keyboards? We should confirm that this works [at least for these keyboards](https://github.com/LightTable/LightTable/issues/620#issuecomment-59621776). 
   * @rundis Could you see if this works for a Norwegian keyboard? Unless I'm mistaken the current behavior doesn't work for you by default
* Currently it's easy for a user to autoclose a new character by adding to the keymap. See [the old keymaps that were used before](https://github.com/LightTable/LightTable/pull/1713/files#diff-e8ac19b4284388a8fd7e980f35d57197). How would a user do this with the closebrackets behaviors?
* `:lt.objs.settings/pair-keymap-diffs` uses the close-pair, open-pair and repeat-pair commands from [auto_parens.cljs](https://github.com/LightTable/LightTable/blob/863a7b289eab2a6d2539865bcad565410a15285f/src/lt/plugins/auto_paren.cljs). Those commands are aware of editor.keys.normal context while the new behavior won't be. Will the new behavior cause any issues for emacs or vim plugins?
* How confusing will it be to have two competing behaviors that autoclose/pair characters?
* If we replace the current behavior, should we remove most of the pair functionality in auto_complete.cljs? 

Once we've answered these questions, we can decide if the closebrackets behavior should be made a temporary alternative, a permanent alternative or a replacement to `:lt.objs.settings/pair-keymap-diffs`. If the closebrackets behavior works for the above international keyboards, we can also consider enabling it by default as LightTable used to before #1713

Currently I'm busy with [0.9.0 issues](https://github.com/LightTable/LightTable/milestones/0.9.0) and won't be able to get to this for some time. If you'd like to move this pull further, I'd recommend looking into and answering what you can of the first 3 questions. @kenny-evitt @rundis Feel free to help with the 3 first three if you're interested

kenny-evitt:
@TeachMeFly I think the first priority for this should be to confirm that it works for the other keyboards we know someone has used in the past.

@cldwalker Is there no way to reliably test this without actually having a (physical) international keyboard?

rundis:
@kenny-evitt Surely you can just change keyboard layout in your os (do that from time to time to get english layout). I'll try out the close brackets thing hopefully tommorow and report back.

kenny-evitt:
@rundis I'm so provincial I've never used a non-US (or even non-QWERTY) keyboard that wasn't a phone keypad (or similar) so I had no idea.

TeachMeFly:
@cldwalker Now it possible to pass option to enable-brackets-autoclose, anyway comdemirror modes know about this addon and with default options repeating single quites will be prevented in clojure and behaviors editors.

1. Now I can only say that it work fine with russian and with [french](https://github.com/LightTable/LightTable/issues/2003#issuecomment-166435391) layouts but I think it will be work everywhere where codemirror can.

2. User can add option to enable behavior, for example ```[:editor.html :lt.objs.editor/enable-brackets-autoclose "[]<>"]``` cause autoclose square and angle brackets in html editors. 

3. I try it with vim plugin, and I have bug, when you open new editor, you can type '[' (and it will be autoclose) until you change mode to insetr first time, after when you back in normal mode it work fine.I didn't test it whith emacs yet but shure it have same problem. I trying to find solution, but it may be better to add some changes to plugin.

4. It possible to enable autoclose-behaviors per editor with tags, but ```auto_parens.cljs``` can be enabled only globally, so it not possible to have own autoclose behavior per editor. Anyway if both of them will be enabled in config it will be work fine unless you still have current autocompleate pairs issue, and new behavior add new issue with vim plugin.

5. No, it just replace ```auto_parens.cljs``` functionality. I think it you must leave ```auto_parens.cljs``` too, for compatability.

So I think new autoclose behaviors would be good alternative for users, who need keyboard different from US, and good replacement in future. Enabling it by default or removing ```auto_parens.cljs``` can cause issues with plugins or breake some user configs.

@kenny-evitt yes, you right.
@rundis I think good way to test all of them inside virtualbox machine, or other, to keep your os clear.

TeachMeFly:
I just tested autoclose-behaviors with:
* German
* Iceland
* Hungary
* Norway
* Italy
* Portuguese

layouts on linux, everything ok!

cldwalker:
> layouts on linux, everything ok!

Great! By everything ok, did you try [all these brackets?](https://github.com/LightTable/LightTable/blob/863a7b289eab2a6d2539865bcad565410a15285f/deploy/core/node_modules/codemirror/addon/edit/closebrackets.js#L12)

cldwalker:
> User can add option to enable behavior, for example [:editor.html :lt.objs.editor/enable-brackets-autoclose "[]<>"] cause autoclose square and angle brackets in html editors.

Great!

> It possible to enable autoclose-behaviors per editor with tags, but auto_parens.cljs can be enabled only globally, so it not possible to have own autoclose behavior per editor.

Good point and definitely a good reason to consider closebrackets.js

> I try it with vim plugin, and I have bug, when you open new editor, you can type '[' (and it will be autoclose) until you change mode to insetr first time, after when you back in normal mode it work fine.I didn't test it whith emacs yet but shure it have same problem. I trying to find solution, but it may be better to add some changes to plugin.

I don't have time to look into this yet but we will need to make sure closebrackets.js works with vim + emacs before we can recommend this solution to other users. Once we resolve this, I would like to make closebrackets.js the autoclose/pairing solution we recommend to users

kenny-evitt:
@TeachMeFly Thanks for checking!

cldwalker:
I've tried all the pairing characters on 1/2 the international keyboards and I'm satisfied. While auto_parens is easy to understand cljs, it's a buggy international support is a deal breaker. closebrackets.js is the way forward for us and I'll be looking to make it the default for this week's release. Hopefully by 0.9.0 we can have it enabled by default. @TeachMeFly thanks for nudging us on this!

I'll throw up a PR tomorow that uses this one and adds a few things - backspace removal, deprecation, etc. As for vim + emacs, I only had issue with Emacs because it's backspace conflicted with closebrackets. I'll cut a release that removes that keybinding from it. @TeachMeFly I wasn't able to reproduce the Vim issue. If you're concerned about it please give me exact steps to reproduce an issue e.g. 1. Press [ 2. Press cw 3. etc

TeachMeFly:
@cldwalker, I've been very busy last week.
You can reproduce issue with vim plugin whith that steps:
    1. Run lighttable;
    2. Open some file;
    3. Put cursor anywhere (but not it start position);
    4. Press [, and you get insterted bracket (and closing bracket) in this position.
I already fix this issue and create [this PR](https://github.com/LightTable/Vim/pull/65) to vim plugin.

Issue #1720, opened 2014-11-16
Modal dialogs overlap each other


can3p:
This report summarize a bunch of tasks connected with prompt dialogs behavior, e.g. #709, #1061, #937 that are closed, but the issue still exists. The main thing is that there is nothing prevents LT from opening new popups over and over again on top of each other, that leads to some ui and logic issues.

Example cases:
1) Open a file in LT and modify it slightly.
2) Open the same file in other editor, then modify and save it multiple times.

On each save LT will produce another dialog about file changes on top of previous one.

Expected behavior: only one prompt dialog at the end

Another case: do the same actions and delete file at the end. LT will show another prompt window above all other that ask about file reload. If you press "ok", this dialog will close, but you will still see
prompts about file changes and if you click "Reload from disk" on any of them, LT will produce JS exception, and buffer will get empty.

Moreover dialogs only look like modal ones, in reality user can do anything with keyboard (e.g. ctrl - space, or even edit the buffer)

cldwalker:
Closing as master no longer has these dialogs for modified or deleted files

PR #2094, opened 2016-01-03
Improve usability with git and any other vcs

### LINKED ISSUE #942: Light table is not git friendly
Switching between branches is really a painful when working with light table. If LT spots
any missing files it's very sad (see #941), then I switch back to a make LT happy again,
but it still thinks files are diff and won't let me do anything unless I close all buffers.

LT should recognize that files are back to normal and they are no longer missing nor changed.

cldwalker:
This fixes #942 and #1720 - usability issues with files being modified and deleted when doing git (or any vcs) operations outside LightTable. Previously, when a file was modified or deleted externally, we would display a popup on each tab and display multiple popups on each tab for each time the file was modified. This was painful as switching or merging branches often led to clicking on a number of dialogs. The new behavior mimics Atom which I think is more friendly and intuitive. The new behavior for a modified file with unsaved changes is to leave the tab as is. As explained in the commit, a user will still get a dialog on save which prevents accidental overwrite. The new behavior for a deleted file is to remove its tab if there are no unsaved changes and to leave the tab as is if there are unsaved changes. @rundis @kenny-evitt I'll merge on Tuesday unless there are concerns

### DIFF
--- a/deploy/settings/default/default.behaviors
+++ b/deploy/settings/default/default.behaviors
@@ -180,7 +180,6 @@
  [:editor :lt.objs.editor.pool/track-active]
  [:editor :lt.objs.editor.pool/ed-close]
  [:editor :lt.objs.editor/menu!]
- [:editor :lt.objs.editor.pool/warn-on-active]
  [:editor :lt.objs.editor/refresh!]
  [:editor :lt.objs.editor.pool/line-comment-options {}]
  [:editor :lt.plugins.auto-paren/repeat-pair]
--- a/src/lt/objs/document.cljs
+++ b/src/lt/objs/document.cljs
@@ -166,7 +166,7 @@
                          (cb d))))))
 
 (defn check-mtime [prev updated]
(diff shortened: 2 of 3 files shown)


kenny-evitt:
I have no concerns, tho I can't easily evaluate the actual changes. But I heartily concur with the spirit and I look forward to the outlined benefits.

cldwalker:
@kenny-evitt Here's some acceptance criteria if you'd like to try it:
* Have a modified FILE (with unsaved changes) open in a tab. `touch FILE` in a terminal. Tab does not throw a popup. Attempting to save will trigger "It looks like this file was modified" popup
* Have a modified FILE open in a tab. `rm FILE` in a terminal. tab stays open. Saving file saves with modified content. Or close tab and file remains removed
* Have an unmodified FILE open in a tab. `rm FILE` in a terminal. tab is removed

I'll merge tomorrow if I don't hear back

Issue #1733, opened 2014-11-20
Refresh tab when trailing whitespaces are removed


integricho:
The behavior `:lt.objs.editor.file/remove-trailing-whitespace` does not refresh the tab where it was executed on save. So even though it removed the trailing whitespaces, the file has to be closed/reopened in order to see that those changes actually happened. Actually, the same goes for `:lt.objs.editor.file/last-char-newline` as well.

cldwalker:
Sorry for the delay. Agree this is a bug. Fix coming shortly

PR #2097, opened 2016-01-06
Fix editor save behaviors

cldwalker:
As pointed out in #1733, there is a bug where the tab doesn't update when trailing whitespace is removed. Fix involved adding a new editor function as CodeMirror's setValue doesn't keep cursor position. Also fixed a related bug where cursor position always jumped when file was modified outside LT. Now there's a fix where cursor position is kept for cases explained in commit. To QA, open a file, move cursor somewhere, `touch FILE` in terminal and observe cursor stay put.

@rundis @kenny-evitt Merging tomorrow night unless there are any concerns

### DIFF
--- a/src/lt/objs/editor.cljs
+++ b/src/lt/objs/editor.cljs
@@ -29,9 +29,18 @@
   [e]
   (.-parentElement (.getScrollerElement (->cm-ed e))))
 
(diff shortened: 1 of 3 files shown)


rundis:
Looks good to me.

PR #2099, opened 2016-01-07
Give user feedback when plugin is (un)installed/updated - close #1920

### LINKED ISSUE #1920: Haskell plugin wasn't listed in installed plugins until after Haskell was installed
I cannot install successfully the Haskell plugin. It says it extracts the plugin, then it says that the behaviors are loaded, but I never actually see it in my installed plugins. I installed the language and 2 seconds later I saw in my installed plugins, so there is clearly some error with it.

cldwalker:
@kenny-evitt I followed up on your idea to make plugin actions less confusing. I also added it to uninstall and update for consistency. I'll merge Friday if I don't hear back. Example screenshot:

screen shot 2016-01-07 at 1 01 17 am

### DIFF
--- a/src/lt/objs/plugins.cljs
+++ b/src/lt/objs/plugins.cljs
@@ -13,6 +13,7 @@
             [lt.objs.deploy :as deploy]
             [lt.objs.notifos :as notifos]
             [lt.objs.tabs :as tabs]
+            [lt.util.js :refer [wait]]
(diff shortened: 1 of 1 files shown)


rundis:
Looks reasonable to me. Feel free to merge

kenny-evitt:
Looks great!

PR #2101, opened 2016-01-08
Make docs navigable by opening in a browser tab - close #1097

### LINKED ISSUE #1097: Forward/Back Buttons for documentation
It would be nice to have a forward/back button for the documentation.  If you follow the links in the official docs there is no way to easily get back to what you were reading.

cldwalker:
Agree that documentation should have navigation so moved it into a browser tab. Removed a bunch of custom code around setting up and styling a custom doc iframe. Documentation no longer opens in a tabset as it was buggy and I don't think provided much value. Before the change:

screen shot 2016-01-08 at 12 23 31 am

After the change - looks the same but is much more usable:

screen shot 2016-01-08 at 12 22 04 am

The doc namespace is just the command but I couldn't think of somewhere else to put it. 

@kenny-evitt @rundis I'll merge Saturday unless there are any concerns

### DIFF
--- a/deploy/core/css/skins/new-dark.css
+++ b/deploy/core/css/skins/new-dark.css
@@ -521,19 +521,6 @@ input {
   background: #606970;
   border: 1px solid #606970;
 }
-.docs ::-webkit-scrollbar {
-  background: #3b3f41 !important;
-}
-.docs ::-webkit-scrollbar-track {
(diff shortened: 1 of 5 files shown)


kenny-evitt:
@cldwalker Looks great!

cldwalker:
Great!

Issue #1460, opened 2014-05-09
Support different fonts for different languages


Reilithion:
I tried to change my user behaviors so that most files show up in the normal Monospace font, but files of a particular language show up in a different font, like so:

    {:+ {:editor [(:lt.objs.style/font-settings "Monospace" "" "")]
         :editor.haskell [(:lt.objs.style/font-settings "Hasklig" "" "")]}

Unfortunately, this seems to have the effect of changing the font everywhere (in all tabs) to the second font, regardless of the language. This happens as soon as a file of that language is opened (I tried it with clojure, too, and it waited until I opened or saved a file with the .clj extension before it applied the new font). Perhaps I am just doing this wrong? But it seems like a bug to me. With this in my user behaviors, I expect Haskell files to use the Hasklig font and all other files to appear in the Monospace font.

cldwalker:
Closing as no one has shown interest in adding this feature. I've added it to [the wishlist](https://github.com/LightTable/LightTable/wiki/Feature-wishlist#general). In the meantime, I'm updating the description from "Editor" to "App" to make it clear it's a global behavior

Issue #1045, opened 2014-01-10
Font settings for a type of editor apply to all editors


puterleat:
I have this in my user behaviours file:

    {:+ {
         ;; The editor tag is applied to all editors
         :editor [:lt.objs.editor/no-wrap
                  (:lt.objs.style/set-theme "default")
                  (:lt.objs.style/font-settings "" "12" "")]

         :editor.python  [(:lt.objs.style/set-theme "neat")
                          (:lt.objs.style/font-settings "" "16" "")]

         }}

But unfortunately all fonts end up 16pt (everywhere). Am I missing something?

cldwalker:
Closing as this is the same as and better explained in #1460

PR #2103, opened 2016-01-09
Font settings is global, not editor specific - #1460

cldwalker:
Users have been mislead in #1460 and #1045 by "Editor:" in the font-settings behavior description. Changing to "App: " makes it clear this is global. Will merge this later today unless there's a concern


Mouvedia:
ref https://github.com/LightTable/LightTable/projects/2#card-166195

PR #2105, opened 2016-01-09
Allow case sensitive renaming - close #1352

### LINKED ISSUE #1352: Changing case when renaming file is not accepted
If I try to rename a file and simply change it's case, lighttable complains and says the file already exists.

This happens if I try to rename
```fileA.js -> FileA.js```

### PR DESCRIPTION
**Pull request by cldwalker**
Latest fs has a rename that is case sensitive. Exists check wasn't case sensitive on osx but able
to work around that by checking if case differs

@rundis @kenny-evitt Will merge later today unless there's a concern

### DIFF
--- a/src/lt/objs/files.cljs
+++ b/src/lt/objs/files.cljs
@@ -243,13 +243,7 @@
 (defn move!
   "Move file or directory to given path"
   [from to]
-  (if (dir? from)
-    (do
-      (.copyDirSyncRecursive wrench from to)
-      (.rmdirSyncRecursive wrench from))
-    (do
-      (save to (:content (open-sync from)))
-      (delete! from))))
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/lt/objs/files.cljs [hunk 1]
```diff
@@ -243,13 +243,7 @@
...
   "Move file or directory to given path"
   [from to]
-  (if (dir? from)
-    (do
-      (.copyDirSyncRecursive wrench from to)
-      (.rmdirSyncRecursive wrench from))
-    (do
-      (save to (:content (open-sync from)))
-      (delete! from))))
+  (.renameSync fs from to))
```
rundis: Much nicer indeed !

rundis:
Looks fine.

cldwalker:
Great!

PR #2076, opened 2015-12-14
use `which` for npm and lein check

ryanpcmcquen:
Signed-off-by: Ryan P.C. McQuen <ryan.q@linux.com>

PR #2075, opened 2015-12-13
make npm check faster/simpler

ryanpcmcquen:
This will need to be verified as working on Windows (as I don't have access to it).

Signed-off-by: Ryan P.C. McQuen <ryan.q@linux.com>

PR #2115, opened 2016-01-15
Quote `which` command in build script

kenny-evitt:
This fixes the Bash file existence check so it handles paths with spaces on Windows.

### DIFF
--- a/script/build.sh
+++ b/script/build.sh
@@ -8,13 +8,13 @@ set -e
 # the app before calling build-app.sh to build it.
 
 # Check if lein is installed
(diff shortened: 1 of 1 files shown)


cldwalker:
I'm guessing this is a different computer than the one QAed for #2075. Could you add the same change for the same lein check a few lines above? Both lein and npm changes were made in #2076

kenny-evitt:
@cldwalker Honestly I'm not sure where I tested it for #2075. Maybe I installed a newer version of npm since then? Anyways, I added the same change for the `lein` check.

I'm going to have to start including which computer I'm working on in these issues!

cldwalker:
:) Looks good!

Issue #1797, opened 2015-01-02
Update codemirror to latest e.g. 5.x


justintaft:
http://codemirror.net/doc/releases.html
Bunch of stuff added since 4.7 (new language support,new vim functionality, etc.)
I'm particularly interested in the new api function getLineTokens.

cldwalker:
This would be nice. In #1772 I looked into upgrading into 4.8 but ran into some errors at startup. I didn't look into them much but I suspect they are due to [key changes](http://codemirror.net/doc/manual.html#normalizeKeyMap). Once this is resolved, we'll need to make sure that the Vim plugin works with a 4.10 codemirror.js and isn't blocked by LightTable/Vim#45. Discussion and pull requests that resolve these issues are welcome

justintaft:
Ah, that explains the funky codemirror directory setup. I'll re-checkout master and take a shot at it.

cldwalker:
Cool. Just `npm install codemirror@4.10` from deploy/core to get started. To be clear, I'm only looking for a dicussion/pull request that can resolve the above issues. If this ends up being difficult, feel free to try #1796 which is an easier codemirror-related issue that would benefit LightTable nicely

justintaft:
Not done yet, but I got Light Table running with 4.10. Able to open up files and change contents and use insta-repl. 

https://github.com/justintaft/LightTable/commit/080d36e2acc4bc317bdb6707e2c983268c124bed

Few files failing to load. Should be simple fix.
LightTableMain/deploy/core/node_modules/codemirror/show-hint.js
LightTableMain/deploy/core/node_modules/codemirror/matchbracket.js
LightTableMain/deploy/core/node_modules/codemirror/search.js

One file was deleted while upgrading, python-hint.js. Haven't tested if hinting still works with python. File was in CodeMirror's repo at one point...
https://github.com/codemirror/CodeMirror/pull/1132

Multiple Light Table guis open when running. A bit odd...

BenjaminVanRyseghem:
should we close this one and focus on #1802 ?

PR #2123, opened 2016-01-20
Autoclose brackets

cldwalker:
This builds off of and supersedes #2093. I've tried 1/2 the international keyboards and am satisfied with intl support. Since fixing backspace in 752d1f1 I haven't seen any issue with backspace in Emacs so we are good as far as emacs and vim. With the closebrackets.js I do see different defaults depending on the filetype. For example, js autocloses '[{(", clj/cljs autocloses [{(" and plain text autocloses [{(. I'm fine with these defaults for now, the clj/cljs matching what we currently have. It will be nice with this behavior to have file type specific pairing.

I've put in a deprecating warning for pair-keymap-diffs which will show up whenever a behaviors file is saved. Here's the current plan I'm thinking for autoclose-brackets and pair-keymap-diffs:

0.8.1 - Make autoclose-brackets the recommended default. Deprecate pair-keymap-diffs. We can reverse this if we find any showstopper autoclose-brackets issues.
0.8.x - Work out any kinks with autoclose-brackets including upgrading CM #1797
0.9.0 - Enable autoclose-brackets by default and remove pair-keymap-diffs

There is an open question of what to do with [these commands](https://github.com/LightTable/LightTable/blob/master/src/lt/plugins/auto_paren.cljs#L83-L105). If autoclose-brackets works for most ppl I'm inclined to deprecate them in 0.8.2 since autoclose-brackets can do anything those commands can. There is a slight difference in that the commands add to existing autoclose characters while `autoclose-brackets` requires you to list all characters. We could an add-autoclose-bracket behavior if users need it.

@kenny-evitt @rundis @TeachMeFly I'll merge tomorrow night unless there are concerns


rundis:
Looks good. (Haven't had a chance to try it out). 
Like the deprecation warning :)

Issue #1699, opened 2014-11-05
Deprecate, replace or remove old functions


cldwalker:
This is _not_ a high priority task but publicly tracking so contributors are aware of this. This was brought up in #1698. Please don't work on this until we've scheduled it for a release. We will want to deprecate and/or remove the following unused functions/behaviors (keeping list up to date):
- [x] `lt.util.cljs/js->clj` ships with clojurescript
- [x] `lt.objs.titlebar/add-titlebar` - looks bad and is not enabled by default
- [x] Remove [unused structure styling](https://github.com/LightTable/LightTable/blob/306a6796249e59d54797e6af5e5331efae7a211f/deploy/core/css/structure.css#L322-L350)
- [ ] `lt.objs.sidebar.command/exec!` and `lt.objs.sidebar.command/command` - We should replace all references to these with their `lt.objs.command` equivalents and then remove them. JavaScript and Emacs plugins have references to them
- [ ] `lt.objs.clients/->name`
- [ ] `lt.ojs.context/group!`
- [ ] `lt.objs.document/replace` - duplicate of one in editor.cljs
- [ ] `statusbar/arrow-disp`
- [ ] `proc/test-printer`
- [ ] `lt.objs.eval/try-read`
- [ ] `:lt.object/tags-removed` triggers in editor behaviors: While looking at #2093 I noticed each time a tab is changed which seems wasteful
- [ ] Move [instarepl styling](https://github.com/LightTable/LightTable/blob/306a6796249e59d54797e6af5e5331efae7a211f/deploy/core/css/structure.css#L118-L128) to instarepl plugin

What to do when we deprecate is up for discussion. It's also possible some functions can be removed. We should also consider at some point marking what namespaces have public APIs. While behaviors are the public apis plugin developers normally hook into, there are namespaces they depend on as well e.g. editor/*, pool/last-active, cmd/command, etc.

EDIT: Be sure a [LightTable github search](https://github.com/search?utf8=%E2%9C%93&q=proc%2Fexec+user%3ALightTable&type=Code&ref=searchresults) doesn't turn up anything for these fns

PR #2124, opened 2016-01-21
0.8.1 cleanup

cldwalker:
This addresses the first three items in #1699. For the titlebar, here's a screenshot of what it looks like:
screen shot 2016-01-21 at 12 39 03 am

In the upper left you can see tiny buttons. As you can see, this was just a placeholder until LT got node-webkit desktop functionality. I only found one plugin that ever reffered to titlebar and it was referring to node-webkit objects that no longer exist in the ns so it was already broken before this PR

Going out with tomorrow's release unless there are concerns


rundis:
Cool. Always nice to nuke unused stuff.

kenny-evitt:
This looks good.

Issue #2038, opened 2015-11-08
Tabset is merged to the left on resize


geusebi:
A tabset is merged on the left whenever I try to resize it horizontally, *BUT* it's happening only on non-primary displays.

Behaviour first seen on:
Light Table version 0.8.0-alpha
Binary version 0.31.1

Wasn't happening on:
Light Table version 0.7.2
Binary version 0.8.6

Three-headed display with primary display located on the left
OS: Linux 4.2.0-1-amd64 SMP Debian stretch 4.2.5-1 (2015-10-27) x86_64 GNU/Linux
Graphics card: GeForce GTX 750 Ti
NVIDIA Driver Version: 340.93

[Example image](https://www.dropbox.com/s/d4b9rd4l4vboo0p/lighttable-tabset.png?dl=0)

geusebi:
If it can help I've noticed this (best explained with an animation):
![anim](https://cloud.githubusercontent.com/assets/1700752/11050598/cd677d92-8747-11e5-9547-2474509a8196.gif)

Apparently the distance between the left margin of the window and the left margin of display is erroneously added the the expected position; and on overflow the tabsets are merged.

Trapets:
Also on Win7 Pro (64 bit), LT v 0.7.2 (binary v 0.11.1-pre).

kenny-evitt:
I was able to reproduce this just now with a recent source build (more recent than 0.8.0-alpha) on a Windows 7 computer.

I was able to confirm, via `lt.objs.console.log` calls, that the `MouseEvent` object being passed to the reaction function for [the behavior `::set-width-final!`](https://github.com/LightTable/LightTable/blob/master/src/lt/objs/tabs.cljs#L429-L461) in *src/lt/objs/tabs.cljs* has a `clientX` property that's greater than the same property for the supplied `MouseEvent` object passed to the reaction function for [the behavior `::width!`](https://github.com/LightTable/LightTable/blob/master/src/lt/objs/tabs.cljs#L463-L492).

Here are the relevant `MouseEvent` properties, logged via `(.log js/console e)`. The final event of the `::width!` behavior reaction function:

```
altKey: false
bubbles: true
button: 0
buttons: 1
cancelBubble: false
cancelable: true
charCode: 0
clientX: 871
clientY: 374
ctrlKey: false
currentTarget: null
dataTransfer: DataTransfer
defaultPrevented: false
detail: 0
eventPhase: 0
fromElement: null
keyCode: 0
layerX: 8
[...]

PR #2147, opened 2016-02-22
Remove set-width-final! see #2038

geusebi:
This fix seems to solve #2038 but I actually don't know why or how it looks to fix it :)
It needs testing before being considered for a merge.

### DIFF
--- a/src/lt/objs/tabs.cljs
+++ b/src/lt/objs/tabs.cljs
@@ -428,40 +428,6 @@
                       (dom/remove-class (dom/$ :body) :dragging)
                       ))
 
-(behavior ::set-width-final!
-          :triggers #{:end-drag}
-          :reaction (fn [this e]
-                      (when-let [ts (next-tabset this)]
-                        (let [width (dom/width (object/->content multi))
-                              left (:left @multi)
-                              cx (.-clientX e)
-                              new-loc (- (+ width left) cx)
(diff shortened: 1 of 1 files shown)


sbauer322:
This fix seems to address the issue and I have not found any problems introduced by removing the code.

@cldwalker, @rundis - any interest in testing this out as well, or are we clear to merge?

cldwalker:
@sbauer322 Before removing a behavior as a fix, it'd be good to dig in and understand what the behavior is providing and for what cases it works and doesn't work (I haven't written this and don't have context). If we are unable to fix it, we should determine if removing it from the codebase will have a negative effect on some users. If it will, then an alternative could be to have not it enabled by default e.g. removing https://github.com/LightTable/LightTable/blob/acf9514f022b61ee2147416e40e81e45ce3f9454/deploy/settings/default/default.behaviors#L556. Kenny digged into this some at https://github.com/LightTable/LightTable/issues/2038#issuecomment-157209962

sbauer322:
> we should determine if removing it from the codebase will have a negative effect on some users

@cldwalker - could you elaborate on how we would test this with users?

cldwalker:
> @cldwalker - could you elaborate on how we would go about testing this with users?

I'm not suggesting testing this w/ users but rather confirming that this works locally for the different scenarios mentioned in #2038 - different # of monitors and possibly on different platforms

sbauer322:
Thanks for clarifying!

PR #2164, opened 2016-03-18
Add menu for underline-result - close #2161

### LINKED ISSUE #2161: missing menu! behavior for underline-result object
hey guys,
I was updating the plugin that I published to use the :inline.underline-result object and in the process I found out that it lacks the menu! behavior. Thus an underline result can only be cleared from the lateral control panel `Clear inline results`.

The relevant lines are [these](https://github.com/LightTable/LightTable/blob/53a90bee4c2c6758e4b5dffdffbc32256a05e00d/deploy/settings/default/default.behaviors#L425-L427):

```
[:inline.underline-result :lt.objs.eval/result-menu+]
[:inline.underline-result :lt.objs.eval/ex-clear]
[:inline.underline-result :lt.objs.eval/destroy-on-cleared]
``` 
I tested it and simply adding  `[:inline.underline-result :lt.objs.menu/menu!]` fixes the problem. 
Note: I guess the underline-result object is sort of like a zombie object since I think it is barely used, thus this never poped up.

Hope it helps

cldwalker:
Also implemented copy for doc underline-result since menu has copy + result
but copy wasn't implemented

Will merge on Saturday unless there are any concerns

### DIFF
--- a/deploy/settings/default/default.behaviors
+++ b/deploy/settings/default/default.behaviors
@@ -425,6 +425,8 @@
  [:inline.underline-result :lt.objs.eval/result-menu+]
  [:inline.underline-result :lt.objs.eval/ex-clear]
  [:inline.underline-result :lt.objs.eval/destroy-on-cleared]
+ [:inline.underline-result :lt.objs.menu/menu!]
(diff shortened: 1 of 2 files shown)


kenny-evitt:
:+1:

PR #2168, opened 2016-03-25
Add the requirement of wget to the cygwin install instructions

ZombieHippie:
`wget` is required to run the install, so I've added that to the cygwin install section.


kenny-evitt:
@ZombieHippie Good catch. I just started setting-up a new Windows computer [L] for working with Light Table and I ran into this, but not in a way that made me think to update the developer install document. So, given that, I'm going to quibble with your changes.

*wget* isn't necessary for Light Table itself. I imagine that you ran into the same thing I did – installing Leiningen using the 'self-installing' Bash script failed because *wget* wasn't installed.

It's nice, for the Light Table maintainers, to limit the info in this document. What you're proposing is adding instructions on installing the Light Table development dependencies themselves. I sympathize with the idea, but I don't think it's ultimately a good idea; mainly because it would require us to maintain those instructions subject to changes in the dependencies themselves. It's also a nice to be able to *assume* that anyone interested in working on Light Table is capable of installing the dependencies themselves.

This would probably be more appropriate as a pull request for the "Install" instructions on the Leinengen site itself.

ZombieHippie:
Good point! It was no trouble to propose.

cldwalker:
Agree with @kenny-evitt. Thanks for the proposal though

Issue #2138, opened 2016-01-31
Upgrading isn't working


kenny-evitt:
There are several recent bug reports for users that upgraded in LT, for 0.8.1 in particular, that are reported as being fixed by downloading the relevant release package directly. Some of those issues:

 - [Error: incorrect header check at Zlib._binding.onerror (zlib.js:295:17) · Issue #2130 · LightTable/LightTable](https://github.com/LightTable/LightTable/issues/2130) – this involved (at least in two cases) upgrading from 0.6.7 and 0.7.2 to 0.8.1
 - [Docs window is to small · Issue #2131 · LightTable/LightTable](https://github.com/LightTable/LightTable/issues/2131)
 - [Button - Exit Fullscreen is floated to left side · Issue #2135 · LightTable/LightTable](https://github.com/LightTable/LightTable/issues/2135)
 - [LT "updates" to 0.8.1 on every restart · Issue #2136 · LightTable/LightTable](https://github.com/LightTable/LightTable/issues/2136)
 - [autoclose-brackets doesn't work in v0.8.1 · Issue #2137 · LightTable/LightTable](https://github.com/LightTable/LightTable/issues/2137)

[I was able to reproduce one of those bugs on a Windows 10 computer [L] after upgrading in LT from 0.8.0 to 0.8.1.](https://github.com/LightTable/LightTable/issues/2131#issuecomment-175963927) I downloaded the release package for 0.8.1 directly and compared the directories where the 0.8.0 and 0.8.1 versions were installed; [a lot (all?) of the 0.8.0 version files seem to not have been upgraded](https://github.com/LightTable/LightTable/issues/2131#issuecomment-175970266).

[As explained previously](https://github.com/LightTable/LightTable/issues/2130#issuecomment-174851856), too much has changed between versions 0.7.2 and 0.8.0 so this should be restricted to upgrades between 0.8.0 and 0.8.1 (or newer verions).

kenny-evitt:
I was able to reproduce #2131 just now after downloading and install 0.8.0 and then supposedly upgrading it to 0.8.1.

kenny-evitt:
So, to clarify, I was able to reproduce this issue just now on a Windows 7 computer [T].

kenny-evitt:
[...]

Issue #2136, opened 2016-01-28
LT "updates" to 0.8.1 on every restart


SnoringFrog:
Every time I restart, LT claims it can update to 0.8.1. I say 'dowload and install'; it tells me the install was successful; I close it and restart; the cycle starts over. 

Each time I get these errors in the console:

```
Error: ENOENT: no such file or directory, scandir '/deploy/'
    at Error (native)
    at Object.fs.readdirSync (fs.js:813:18)
    at Object.fs.readdirSync (ATOM_SHELL_ASAR.js:454:28)
    at lt.objs.files.full_path_ls (file:///Users/430011037/Applications/LightTable.app/Contents/Resources/app/core/node_modules/lighttable/bootstrap.js:21994:134)
    at lt.objs.deploy.move_tmp (file:///Users/430011037/Applications/LightTable.app/Contents/Resources/app/core/node_modules/lighttable/bootstrap.js:28969:215)
    at Extract.<anonymous> (file:///Users/430011037/Applications/LightTable.app/Contents/Resources/app/core/node_modules/lighttable/bootstrap.js:28988:31)
    at emitNone (events.js:72:20)
    at Extract.emit (events.js:166:7)
    at DirWriter.<anonymous> (/Users/430011037/Applications/LightTable.app/Contents/Resources/app/core/node_modules/tar/lib/extract.js:66:8)
    at emitNone (events.js:72:20)

Error: ENOENT: no such file or directory, scandir '/Users/430011037/Applications/LightTable.app/Contents/Resources/app/tmp/'
    at Error (native)
    at Object.fs.readdirSync (fs.js:813:18)
    at Object.fs.readdirSync (ATOM_SHELL_ASAR.js:454:28)
    at lt.objs.files.full_path_ls (file:///Users/430011037/Applications/LightTable.app/Contents/Resources/app/core/node_modules/lighttable/bootstrap.js:21994:134)
    at lt.objs.deploy.move_tmp (file:///Users/430011037/Applications/LightTable.app/Contents/Resources/app/core/node_modules/lighttable/bootstrap.js:28969:70)
    at Extract.<anonymous> (file:///Users/430011037/Applications/LightTable.app/Contents/Resources/app/core/node_modules/lighttable/bootstrap.js:28988:31)
    at emitNone (events.js:72:20)
    at Extract.emit (events.js:166:7)
[...]

Issue #2131, opened 2016-01-22
Docs window is to small


ProjectINT:
![screenshot from 2016-01-22 08 52 37](https://cloud.githubusercontent.com/assets/16366604/12504407/e2ae6cd8-c0e5-11e5-9b17-0fb10f66214a.png)

cldwalker:
Thanks for the report. That looks like an old version. Upgrading to 0.8.1 should fix this. You can also just moving the docs tab next to Welcome tab and close the empty tabset. Closing but if you have trouble the upgrade please give more info as asked in CONTRIBUTING.md

kenny-evitt:
[Someone else reported this on Gitter.](https://gitter.im/LightTable/LightTable?at=56a77c8ceaf741c118d4f384)

But I couldn't reproduce this myself just now.

@ProjectINT What steps are you taking that result in the docs tab appearing like in your screenshot?

Tuborgman:
Hello! This is reproducible on OSX OSX 10.11.3.
Steps are simple:
1. Open the editor.
2. Click on the highlighted link "Light Table's online docs"
3. Observe the issue on the just opened right pane.
http://pasteboard.co/15ckPBxw.png

kenny-evitt:
I can't reproduce this [note to me – T] on a Windows 7 computer running a recent source build (at least as recent as version 0.8.1).

Interestingly, when I *left* click on the link, a new tab is opened – in the same tabset as the "Welcome" tab.

The docs tab is also named differently – my tab is named "Light Table docs". But in both of the screenshots provided by @ProjectINT and @Tuborgman, the tab is named "Docs".

@Tuborgman Did you close LT since you upgraded to 0.8.1? Would you try closing it and then confirm whether you can reproduce this afterwards?

kenny-evitt:
I was able to reproduce this on a Windows 10 computer [note to me – L] just now.

I was running 0.8.0. The Docs opened fine – old style.

I upgraded to 0.8.1. When I closed and re-opened LT and then clicked the docs link I got the shrunken docs tab in a new tabset.

kenny-evitt:
I just downloaded 0.8.1 from the website on the same Windows 10 computer. The docs link opens in a tab in the *same* tabset, and it's not shrunk.

kenny-evitt:
[...]

PR #2176, opened 2016-04-04
Fix upgrading

kenny-evitt:
The goal is to download the appropriate release file for each platform. Currently, the source tarball is being downloaded but that doesn't contain any of the compiled ClojureScript code, so nothing is really upgraded.

This is currently definitely a work-in-progress.


kenny-evitt:
@cldwalker @rundis I've basically got this working. It seems to work entirely on Windows but not on Mac (and I haven't even tested at all on Linux yet).

On my Mac, the code to delete the temp files (the *tmp* sub-directory and the actual downloaded archive file) is failing (and silently). I'm debugging it now.

I'm also going to separate the current WIP commit into at least two separate commits – the first with just the new Node modules I added. As-is, there are too many files to view the actual code changes in GitHub!

Once I've got the code working for me on all three platforms, then I'll start cleaning up my changes.

kenny-evitt:
@cldwalker @rundis I've got this working on all platforms. There's definitely room to cleanup the code. But I'd like to get some feedback from both of you – or at least one of you – before I start working on that.

To test, you can checkout [the branch 0.8.0-fix-upgrading-test in my fork](https://github.com/kenny-evitt/LightTable/tree/0.8.0-fix-upgrading-test) and build from source. I cherry-picked my two commits in this PR on top of the code in *master* as-of the 0.8.0 tag.

kenny-evitt:
[Here's the files-changed diffs for just the "WIP" commit](https://github.com/LightTable/LightTable/pull/2176/commits/7867d36060fb7e7ee2fa3c4c34d056aa5c39a918) (as the other commit only adds files for two Node modules I added).

src/lt/objs/deploy.cljs
```diff
@@ -21,31 +21,64 @@
...
   (let [vstr (:content (files/open-sync (files/lt-home "core/version.json")))]
     (js->clj (.parse js/JSON vstr) :keywordize-keys true)))
 
-(defn proxy? []
-  (let [p (get-proxy (tar-path "0.5.0"))]
-    (when (str-contains? p "PROXY")
-      (-> p
-          (string/split " ")
-          (second)))))
+;; (defn proxy? []
```
kenny-evitt: I commented this out because it calls `tar-path`, but that now requires a callback, and I couldn't find any callers of this function anyways.

src/lt/objs/deploy.cljs
```diff
@@ -85,38 +118,145 @@
...
+;(download-zip "0.8.1" (fn [e r body] nil))
 
 (defn untar [from to cb]
   (let [t (.createReadStream fs from)]
     (.. t
         (pipe (.createGunzip zlib))
-        (pipe (.Extract tar (js-obj "path" to)))
-        (on "end" cb))))
+        (pipe (.extract tar to (js-obj "fs" original-fs)))
+        (on "finish" cb))))
```
kenny-evitt: The *tar-fs* library, unlike the *tar* library, doesn't fire an *end* event.

kenny-evitt:
The shitty change – that I can't figure out how to avoid without even shittier (or much harder) changes – is that I had to explicitly handle *.asar* files in our (previously) minimally modified shell.js Node module fork.

kenny-evitt:
@cldwalker @rundis This could definitely use a thorough review, ideally from both of you.

carocad:
@kenny-evitt: I build LT from the 0.8.0 tag cherry picking the commit WIP and although it correctly builds, I get an error when starting LT. Here is what I've got:

```
Error: Cannot find module '/home/camilo/Proyectos/LightTable/builds/lighttable-0.8.0-linux/resources/app/core/../core/node_modules/tar-fs'
    at Function.Module._resolveFilename (module.js:336:15)
    at Function.Module._load (module.js:286:25)
    at Module.require (module.js:365:17)
    at require (module.js:384:17)
    at lt.util.load.node_module (file:///home/camilo/Proyectos/LightTable/builds/lighttable-0.8.0-linux/resources/app/core/node_modules/lighttable/bootstrap.js:18832:10)
    at file:///home/camilo/Proyectos/LightTable/builds/lighttable-0.8.0-linux/resources/app/core/node_modules/lighttable/bootstrap.js:28913:47
```

kenny-evitt:
@carocad Thanks! Would you cherry-pick both commits, build, and then try testing the upgrade again please? The error refers to a Node.js package added in the commit prior to the "WIP" commit.

carocad:
@kenny-evitt I did as you said. The build process succeds without any problem, LT starts without a problem.
The pop-up requesting me to update LT does work. When I click it I get a text in the console with a git... LT...0.8.1 tar-file download logged out.

But it all ends there :(
After that, there are no more logs, windows or anything like that. I checked the CHANGELOG and it is still 0.8.0, which is quite weird because my pre-built LT did upgraded correctly, so no idea.
Another weird thing is that it requested me to download the javascript plugin which I already had.

A final note: Something not so funny now happens on my normal LT. Everytime that I open my normal LT it asks me to install paredit and rainbow, which I purposely deleted before ! 
It doesn't matter if I deny the action, next time that I open it, it is there. If you know a way to revert that I would appreciate it

Hope it helps

kenny-evitt:
@carocad Was the CHANGELOG updated? It should include a section for 0.8.1 at the top. And you should have seen a "Light Table has been updated!" popup when the upgrade finished that suggested you restart LT too.

I'll look into this soonish.

carocad:
@kenny-evitt yes to all your questions. It works as expected

kenny-evitt:
@carocad Oh! Cool, thanks!

martinchooooooo:
Hey @kenny-evitt I tested your changes and all looks good to me!

Here's what I did:
`git clone git@github.com:kenny-evitt/LightTable.git`
`git co 0.8.0-fix-upgrading-test`
`./script/build-app.sh`
`./builds/lighttable-0.8.0-mac/light`

- Was prompted straight away to update to 0.8.1. Agreed
- Was prompted that the following plugin dependencies are missing: Javascript 0.2.0. install All.
- Console printed "https://github.com/LightTable/LightTable/releases/download/0.8.1/lighttable-0.8.1-mac.tar.gz"
- “Light Table has been updated! …”. Ok
- Opened File > About.
- The Changes file shows 0.8.1 with a list of issues, so I guess it went through fine.
- Ran a quick ‘smoke test’ by opening a CLJ file I commonly use, evaluate how file, tried (+ 2 3) and worked fine. Not the most thorough test, but it works :)

#2193 - Can’t reproduce.
#2136 - Can’t reproduce.
#2131 - Can’t reproduce.

OSX 10.11.5, MBPro Late 2011

kenny-evitt:
@martinchooooooo Thanks!

sbauer322:
I was able to test this out as well:

Commands to start off:

```
git clone git@github.com:kenny-evitt/LightTable.git ./kenny-LightTable
cd ./kenny-LightTable
git checkout 0.8.0-fix-upgrading-test
./script/build.sh
./builds/lighttable-0.8.0-linux/light
```

- Immediately prompted to update to 0.8.1.
- Dependencies missing: Clojure 0.3.1, Javascript 0.2.0
- Console output: `https://github.com/LightTable/LightTable/releases/download/0.8.1/lighttable-0.8.1-linux.tar.gz`
- Changelog listed 0.8.1
- Various simple eval-ing of functions worked
- Went about my normal development work and no issues encountered

Excellent work @kenny-evitt

kenny-evitt:
@rundis This is a significant enough change that this needs your approval before it can be merged.

We should also decide how to address or resolve (or not) the various points I mentioned in [this comment](https://github.com/LightTable/LightTable/issues/2138#issuecomment-205104513) on the issue – #2138 – to which this PR corresponds:

> The major change is that we want to download a release file (with the compiled ClojureScript) instead of the source tarball (as the code does now). That's mildly tricky for a few reasons:
> 
>  1. There's a separate release for each platform. In the interest of potentially being able to also upgrade Electron itself, I figured we might as well download the relevant release file for the current user.
>  2. I'm retrieving the release info from GitHub to retrieve the URLs for the release files so there's a new callback now that propagates thru a few functions just to get that URL. Previously, the URL was very simple to generate.
>  3. The Windows release file is a ZIP file so I needed to add a Node module to extract the files. I'm not sure how much we should attempt to avoid this. We *could* publish *.tar.gz* release files for Windows because archive utilities, e.g. 7-Zip, can handle them just fine.
>  4. [Electron patches `fs` to make *.asar* files act as directories](https://github.com/atom/electron/issues/1658) – this breaks both of the ZIP libraries I tried, probably breaks all ZIP libraries, and it *also* apparently breaks tar files (tho I haven't confirmed this myself yet). I couldn't get any of [the seeming workarounds documented by Electron for forcing Electron to treat an ASAR archive as a normal file](https://github.com/atom/electron/blob/master/docs/tutorial/application-packaging.md#treating-asar-archive-as-normal-file) to work for me. [Others have mentioned needing to use libraries that allow an alternate `fs` module to be supplied.](https://github.com/atom/electron/issues/1658#issuecomment-126071165) We can avoid this indefinitely because we don't need to extract the entirety of any of the release archive files – we seem to only need the *lighttable-version-platform/resources/app/* directory and the *.asar* file is in the *lighttable-version-platform/resources/* directory. But then we can't really do Electron upgrades per [1].
> 
> Related to [2], there's an existing 'edge' setting that seems to control whether the user is running the latest *source* version. I'm going to defer implementing that for now, but we could potentially use the Travis CI builds to support that.
> 
> As-is, I'm going to continue with my current 'solution', i.e. by modifying `fetch-and-deploy` to call `unzip` if the platform of the current computer is Windows.

I hazily remember @cldwalker bringing up something about the Node.js packages. I think the first commit in the PR not only adds two packages I used explicitly, but it also included other packages added when I ran `npm install`. I am very much un-interested in that myself but I'm willing to make changes if anyone else has strong feelings about it.

I also need to do the following before this merged:

 - [ ] Cleanup the 'WIP' commit.
 - [ ] Rebase *master* onto this topic branch.

kenny-evitt:
@sbauer322 Thanks for testing this!

src/lt/objs/deploy.cljs
```diff
@@ -21,31 +21,64 @@
...
+(def tar-extension (if (platform/win?) ".zip" ".tar.gz"))
+
+(defn get-release-download-url
+  "Retrieve the download URL for the specified release version and this computer's platform."
+  [v cb]
+  (let [release-info-url (str "https://api.github.com/repos/LightTable/LightTable/releases/tags/"
+                              v)]
+    (fetch/xhr release-info-url
+               {}
+               (fn [data]
```
rundis: That's a whole lotta' lets !
How about something like :
```
(fn [data]
  (->> (-> (js/JSON.parse data)
           (js->clj :keywordize-keys true))
       :assets 
       (filter #(< -1 (.indexOf (:name %) (name platform/platform))))
       first
       :browser_download_url
       cb))
```

src/lt/objs/deploy.cljs
```diff
@@ -85,38 +118,145 @@
...
+
+; TODO: Rename to `download-release-file`?
 (defn download-zip [ver cb]
   (let [n (notifos/working (str "Downloading version " ver " .."))]
-    (download-file (tar-path ver) (str home-path "/tmp.tar.gz") (fn [e r body]
-                                                                  (notifos/done-working)
-                                                                  (cb e r body)))))
+    (tar-path ver
+              (fn [url]
+                (console/log url)
```
rundis: Is this supposed to be here ?

src/lt/objs/deploy.cljs
```diff
@@ -21,31 +21,64 @@
...
-    (when (str-contains? p "PROXY")
-      (-> p
-          (string/split " ")
-          (second)))))
+;; (defn proxy? []
+;;   (let [p (get-proxy (tar-path "0.5.0"))]
+;;     (when (str-contains? p "PROXY")
+;;       (-> p
+;;           (string/split " ")
+;;           (second)))))
```
rundis: no biggie, but you don't actually need the parenthesis when threading and the function in question just takes one arg.

rundis: But I see this is not your doing in the first place...

rundis:
@kenny-evitt Maybe this would help with the ASAR stuff ? https://github.com/electron/electron/pull/3641 

It would definitely be good if we were also able to make upgrades to Electron itself. Otherwise that will become another source of bitrot too.

deploy/core/node_modules/shelljs/shell.js
```diff
@@ -389,8 +389,8 @@ function _rm(options, files) {
     // If here, path exists
 
     var stats = fs.statSync(file);
-    // Remove simple file
-    if (stats.isFile()) {
+    // Remove simple file (or Electron package files which 'fs' has been patched by Electron to treat as directories)
```
rundis: Is that still true with https://github.com/electron/electron/pull/3641 ?

deploy/core/node_modules/shelljs/shell.js
```diff
@@ -1612,7 +1612,7 @@ function rmdirSyncRecursive(dir, force) {
     var file = dir + "/" + files[i],
         currFile = fs.lstatSync(file);
 
-    if(currFile.isDirectory()) { // Recursive function back to the beginning
+    if(currFile.isDirectory() && !(file.substr(-5) === '.asar')) { // Recursive function back to the beginning
```
rundis: Ref noAsar flag fix in Electron

rundis:
> _I hazily remember @cldwalker bringing up something about the Node.js packages. I think the first commit in the PR not only adds two packages I used explicitly, but it also included other packages added when I ran npm install. I am very much un-interested in that myself but I'm willing to make changes if anyone else has strong feelings about it._

Well I do care about dependencies. The fewer the better. I suppose I don't have to mention the left-pad saga with NPM ? There is an npm package for checking whether something is an array or object. That's f*ng oneliners. All these deps are increasing our maintenance burden, bloating the distribution and dumbs us down because we don't care enough about what these things do. It's just a matter of adding a thing to our package.json and behold the internet being pulled down.

I'm going O.T. I'm not saying we should write our own un-taring, but do we need that new dep (and all it's transitives). I'm not sure what's wrong with the one we had. Do we need to keep both ?
Was there any alternatives out there without (that many) transitive deps.

src/lt/objs/deploy.cljs
kenny-evitt: I'm not against your suggestion, and it's definitely much more idiomatic Clojure, but it's much harder to read for me. You also subtly changed the behavior; did you deliberately remove using the version (`v`)? Doing so is *probably* (?) okay, as I don't think we even have (or want to provide) a way to upgrade to any version that's not the latest, my version was (probably) intended to explicitly use the version specified.

src/lt/objs/deploy.cljs
kenny-evitt: Well, yes *and* no :smile: 

This is definitely debris leftover from my hacking this together.

I imagine no one will really find the URL being logged to their console to be of much utility so I'll take it out when I get around to cleaning up this commit.

src/lt/objs/deploy.cljs
kenny-evitt: Note also that I'm commenting out this function entirely. [My original comment about these changes](https://github.com/LightTable/LightTable/pull/2176#discussion_r60169719):

> I commented this out because it calls `tar-path`, but that now requires a callback, and I couldn't find any callers of this function anyways.

deploy/core/node_modules/shelljs/shell.js
kenny-evitt: It seems like it's not true with that PR merged. What's weird is that PR was merged on the same date as we last bumped Electron in LT. But we bumped to 0.34.5 and that PR seems to be part of 0.35.3. @sbauer322 might test upgrading to a newer version.

I'll try to do so myself for the changes in this PR too. Given that everyone's going to need to either download the next version or build it themselves from source, upgrading Electron now has no marginal impact. Not having to handle the ASAR workaround should simplify this greatly. Maybe we could even get rid of some other Node.js dependencies.

kenny-evitt:
@rundis Thanks for the comments! I'll need some time to go thru them all; at least more thoroughly.

What I meant about the Node.js packages, specifically, was something to do with them being vendored strangely in the repo. I only added two packages and, yes, their dependencies would be expected to be added to the repo too, but there were also other packages added (I think) that are dependencies of other existing packages. I can't find the original conversation, wherever it was, so I don't remember exactly.

I didn't think the packages had that many dependencies but I just checked out the graph for *tar-fs*:

 - [Visualization of npm dependencies](http://npm.anvaka.com/#/view/2d/tar-fs)

... and – *sigh* – you're right that that looks crazy.

src/lt/objs/deploy.cljs
sbauer322: I do agree this suggestion is harder to read, but feels cleaner once you stare at it for a few minutes. In particular, using the results of `->` as the first argument to `->>` was initially disorientating. 

Maybe use a let binding instead of: 
```
(-> (js/JSON.parse data)
      (js->clj :keywordize-keys true))
```
to improve readability?

Not to self-promote poorly written blog posts (of which I am a fine provider of), but I wrote a rather brief guide about [[arrow macros once](http://bauerspace.com/target-practice-with-the-arrow-macro/).

rundis: @kenny-evitt The v is used for the rest / ajax call, I just didn't bother also using it in the callback traversing the result... as it would be very surprised if we had multiple versioned release artefacts for one release tag.

Both: Yeah I suppose @sbauer322 suggestion is a case where let would help readability.

I'll admit that I use threading quite extensively. Once you get used to them, they read actually quite a lot easier than the normal way to read Clojore/Lisp (inside->out). I think of them like a pipeline which you can just read sequentially.

Anyways I do think that using a let just to set up the following let that sets up the next let etc is a very imperative approach (and all those names doesn't really add that much, because they are just a means to an end - getting that download url. Come to think of it it could be extracted to a separate function) The multiple let approach just doesn't feel particularly functional to me. I guess that was my main point :-)

Issue #2196, opened 2016-05-01
Add behavior to set a GitHub API token to avoid rate limiting for checking for new versions of LT and plugins


kenny-evitt:
[@josser suggested this in the Gitter room](https://gitter.im/LightTable/LightTable?at=5714c4d55ed5a4fd3fe3f000) and linked to a Gist with instructions on how to use the same feature in Homebrew:

 - [Set your Github API Token If you hit a ”GitHub API rate limit exceeded” when searching with Homebrew (http://brew.sh/).](https://gist.github.com/christopheranderton/8644743)

Tho see the comments for some feedback about the instructions in that Gist.

I'm working on this now.

kenny-evitt:
Please un-assign me.

Issue #2065, opened 2015-12-07
GitHub responded with a status of 403


muhuk:
I see this error in my console:

    Failed to load resource: the server responded with a status of 403 (Forbidden): https://api.github.com/repos/LightTable/LightTable/tags

There is also a javascript stacktrace and it is repeating itself.

Lighttable version 0.8.0-alpha, Binary version 0.31.1.

kenny-evitt:
@muhuk Would you do some digging and check whether you might have exceeded [the GitHub API rate limits](https://developer.github.com/v3/#rate-limiting)? You should be able to run `curl -i https://api.github.com/repos/LightTable/LightTable/tags` and see rate limit info in special headers.

muhuk:
@kenny-evitt that's what I've seen when I visited the URL yesterday. Now I can see a list of tags.

    wget -S https://api.github.com/repos/LightTable/LightTable/tags
      ....
      X-RateLimit-Limit: 60
      X-RateLimit-Remaining: 57
      X-RateLimit-Reset: 1449561436

I only had LT open and wasn't using `api.github.com`. Does it check tags every minute?

kenny-evitt:
[It looks like it's running every five minutes](https://github.com/LightTable/LightTable/blob/683785fdf571d32e86aa9fd7e6e98204243207b3/src/lt/objs/deploy.cljs#L188-L198):

```
(behavior ::check-version
          :triggers #{:init}
          :type :user
          :desc "App: Automatically check for updates"
          :reaction (fn [this]
                      ;;                               (when-let [proxy (proxy?)]
                      ;;                                 (.defaults request (clj->js {:proxy proxy})))
                      (when (app/first-window?)
                        (set! js/localStorage.fetchedVersion nil))
                      (check-version)
                      (every version-timeout check-version)))
```

The `every` function accepts a timeout in milliseconds [and](https://github.com/LightTable/LightTable/blob/683785fdf571d32e86aa9fd7e6e98204243207b3/src/lt/objs/deploy.cljs#L50):

```
(def version-timeout (* 5 60 1000))
```

Even tho that should only result in 12 requests an hour, [the GitHub API restricts unauthenticated requests more strictly](https://developer.github.com/v3/#rate-limiting):

> For unauthenticated requests, the rate limit allows you to make up to 60 requests per hour. **Unauthenticated requests are associated with your IP address**, and not the user making requests. 

[...]

PR #1996, opened 2015-09-26
System Tests 

justintaft:
~~DO NOT MERGE. EXPERIMENTAL.~~

Whenever light table runs remote debugging is available. This is a security concern and we may want to change this in the future...
https://github.com/LightTable/LightTable/blob/baa2037aecaf382368ef299ed885414ca0f8e1ac/deploy/core/main.js#L107

To access remote debugging tools:
-  run light table as normal
-  navigate to http://localhost:8315

To debug with web driver:
- run chromedriver -v
- run the  contents of system-tests.clj (binary location will need to be changed)
- Launch light table

Sometimes light table does not initialize all the way. If this happens, close light table & retry.

I would like the system tests to use a cljs or js library, so we don't have to boot up the jvm for testing. Any Thoughts?

Issue #1893, opened 2015-04-24
Having Issues Downloading Any Plugins 


Jaballadares:
Please forgive me if I post too much of the error message. I'm new the world of programming and fear omitting something useful.

Failed to load resource: the server responded with a status of 403 (Forbidden): https://api.github.com/repos/LightTable/plugin-metadata/commits
Uncaught TypeError: Cannot read property 'sha' of undefined
bootstrap.js [29185]	anonymous
bootstrap.js [23098]	anonymous
bootstrap.js [23110]	anonymous
bootstrap.js [14143]	goog.events.Listener.handleEvent
bootstrap.js [14378]	goog.events.fireListener
bootstrap.js [14368]	goog.events.fireListeners_
bootstrap.js [14421]	goog.events.dispatchEvent
bootstrap.js [14542]	goog.events.EventTarget.dispatchEvent
bootstrap.js [16377]	goog.net.XhrIo.dispatchErrors_
bootstrap.js [16408]	goog.net.XhrIo.onReadyStateChangeHelper_
bootstrap.js [16394]	goog.net.XhrIo.onReadyStateChangeEntryPoint_
bootstrap.js [16390]	goog.net.XhrIo.onReadyStateChange_

kenny-evitt:
@Jaballadares Interestingly, your comment in GitHub only seems to show the last block of the error messages that I saw in my email notification for this issue. Did you edit your original comment?

Also, would you please post the steps you can follow to reproduce this and divulge more info about your environment, e.g. OS, LightTable version, Java JRE and JDK versions, Leiningen plugins installed, etc.

kenny-evitt:
@Jaballadares Re-open when or if you have some more details about the problem(s) you're having.

PR #2198, opened 2016-05-01
Use user user-name and token for GitHub API calls

kenny-evitt:
Implements #2196

@cldwalker @rundis How's this look?

### DIFF
--- a/src/lt/objs/deploy.cljs
+++ b/src/lt/objs/deploy.cljs
@@ -11,6 +11,7 @@
             [lt.objs.sidebar.command :as cmd]
             [lt.objs.console :as console]
             [lt.objs.app :as app]
+            [lt.objs.github :as github]
             [lt.util.load :as load]
             [lt.util.js :refer [every]]
             [lt.util.cljs :refer [str-contains?]]
@@ -31,8 +32,8 @@
 
 (defn tar-path [v]
   (if (cache/fetch :edge)
-    (str "https://api.github.com/repos/LightTable/LightTable/tarball/master")
-    (str "https://api.github.com/repos/LightTable/LightTable/tarball/" v)))
+    (str "https://" (github/basic-auth-hostname-prefix) "api.github.com/repos/LightTable/LightTable/tarball/master")
+    (str "https://" (github/basic-auth-hostname-prefix) "api.github.com/repos/LightTable/LightTable/tarball/" v)))
 
 (def version-regex #"^\d+\.\d+\.\d+(-.*)?$")
 
@@ -118,7 +119,11 @@
                                                          restart to get the latest and greatest.")
                                               :buttons [{:label "ok"}]}))))))
 
-(def tags-url "https://api.github.com/repos/LightTable/LightTable/tags")
+(defn tags-url
+  []
+  (str "https://"
+       (github/basic-auth-hostname-prefix)
+       "api.github.com/repos/LightTable/LightTable/tags"))
 
 (defn should-update-popup [data]
   (popup/popup! {:header "There's a newer version of Light Table!"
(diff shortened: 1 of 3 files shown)


cldwalker:
@kenny-evitt Before talking about the code, I'd like to better understand the motivation. I haven't seen this in issues (though I've been less active) and I haven't encountered this issue myself. If this is for one user on gitter and they haven't taken the time to make their case for it on github, I don't think we should be immediately prioritizing it to build. This feature is also somewhat technical e.g. getting an auth token so I don't expect this to be a common user workflow. Probably worth checking if there are other ways to auth besides the url which would then require no change. If there are none, then I'd suggest minimal LT changes to allow GH credentials.

kenny-evitt:
@cldwalker There are a few issues that mention this or seem to involve it:

 - [GitHub responded with a status of 403 · Issue #2065 · LightTable/LightTable](https://github.com/LightTable/LightTable/issues/2065)
 - [System Tests by justintaft · Pull Request #1996 · LightTable/LightTable](https://github.com/LightTable/LightTable/pull/1996)
 - [Having Issues Downloading Any Plugins · Issue #1893 · LightTable/LightTable](https://github.com/LightTable/LightTable/issues/1893)

I created #2196 for the user that mentioned it in Gitter and I was going to point to similar changes, e.g. my adding the Node.js start options behavior, so anyone could work on it. Then I decided 'fuck it, I'll just implement it myself; it shouldn't take long' – and thus this PR. I wasn't prioritizing it as much as trying to kill it off before even adding it to our set of issues. Unfairly of me, I didn't really account for the cost of you or Magnus having to review it – I'm sorry for not doing that.

I also don't expect this to be commonly used. But for anyone that is running into it, there's really no workaround available. If their public IP address is being shared (as I'd expect it is for almost everyone), then it's possible they could run into the limit pretty quickly.

#### The Technology

[There are only three ways to authenticate with GitHub](https://developer.github.com/v3/#authentication):

 1. Basic Authentication
 2a. OAuth2 Token (sent in a header)
 2b. OAuth2 Token (sent as a parameter)
 3. OAuth2 Key/Secret

[1], [2a], and [3] involve modifying the URL, which you want to avoid, but [2b] requires adding a header to all of the GitHub API requests, so I don't understand how that helps us "require no change".

I also can't understand how we could (properly) use OAuth because there's no way for us to keep the client secret (that is generated by GitHub when we register our application with them) secret.

[The GitHub docs](https://developer.github.com/v3/#authentication) do state:

> ... OAuth2 tokens can be [acquired programmatically](https://developer.github.com/v3/oauth_authorizations/#create-a-new-authorization), for applications that are not websites.

But I still don't think there's any way for us to get an authentication token securely. The link in the quoted text above starts:

> If you need a small number of tokens, implementing the web flow can be cumbersome. Instead, tokens can be created using the OAuth Authorizations API using Basic Authentication. To create tokens for a particular OAuth application, you must provide its client ID and secret, found on the OAuth application settings page, linked from your OAuth applications listing on GitHub. If your OAuth application intends to create multiple tokens for one user you should use fingerprint to differentiate between them. OAuth tokens can also be created through the web UI via the Personal access tokens settings. Read more about these tokens on the GitHub Help page.

So, we can create OAuth tokens for our not-a-website application, but we have to use Basic Authentication to do so. I'm also unclear about what credentials we're expected to or supposed to use. Whatever credentials we use, we can't really secure them if they're distributed with the LT app code.

This blog post describes authenticating via OAuth in an Electron app too:

 - [Electron. oAuth authentication with GitHub](http://iamemmanouil.com/blog/electron-oauth-with-github/)

But the author admits that the GitHub application client secret is stored in the Electron app *.asar* package (and that's not secure).

I can't figure out how we could securely authenticate with OAuth. I asked a question on Stack Overflow about it too:

 - [(GitHub) API OAuth authentication for not-a-website applications? - Stack Overflow](http://stackoverflow.com/questions/36989964/github-api-oauth-authentication-for-not-a-website-applications)

It's very possible I'm an idiot and I'm missing something about how it's supposed to work or capable of working.

PR #2205, opened 2016-05-13
Use OS X native tools for CLI command

anddam:
Use metadata store to find the application; drop unnecessary use of bash; rely on system's tool by dropping env.

### DIFF
--- a/deploy/platform/mac/light
+++ b/deploy/platform/mac/light
@@ -1,22 +1,16 @@
-#!/usr/bin/env bash
+#!/bin/sh
 
 if [ -z "$LT_HOME" ]
 then
-	myDir=$( cd $(dirname $0) ; pwd -P )
-	myAppDir="$myDir/../Applications"
-	for i in ~/Applications ~/Applications/LightTable $myDir $myDir/LightTable $myAppDir $myAppDir/LightTable /Applications /Applications/LightTable /Applications/Utilities /Applications/Utilities/LightTable; do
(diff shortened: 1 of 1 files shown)


kenny-evitt:
@anddam Thanks for the PR! This is interesting. I'll try to confirm that this works on my Mac soonish.

Any reason why this is useful or necessary? Do any of the recent Mac OS X versions not include Bash?

But based on the info you've provided, this seems sensible.

anddam:
I downloaded LightTable, had a look at the cli script and:

  - I didn't like the cd;pwd dance, I felt mdfind is easier;
  - I go with sh over bash, unless an actual bash-specific feature is required (not the case);
  - I think there's no need for using any user's custom bash setup via env as the script just want to start the Electron process.

This change is not necessary, it's just a suggestion for a tidier script.

cldwalker:
Thanks for the PR but this breaks our local development: 
```
$ mdfind kMDItemCFBundleIdentifier == com.kodowa.LightTable
/Applications/LightTable.app
/Users/me/code/repo/LightTable/builds/lighttable-0.8.1-mac/LightTable.app
```
The order of prioritized directories in the current script isn't the same as mdfind so it doesn't make sense to use it. In general we aren't looking for contributions to deploy/ or script/ as only the core team should be updating these directories. Apologies for not making this policy more explicit. Will update CONTRIBUTING

anddam:
I see the point, thanks for letting me now.

Now let's focus on this interesting editor of yours ;-)

Issue #1984, opened 2015-09-19
Add back proxy support


cldwalker:
Proxy support [was disabled](https://github.com/LightTable/LightTable/blob/dec215a2846644e5544c971c4275f6d8313ad730/src/lt/objs/deploy.cljs#L193) and not implemented in the move from node-webkit to electron. Updating [the old proxy](https://github.com/LightTable/LightTable/blob/dec215a2846644e5544c971c4275f6d8313ad730/src/lt/objs/deploy.cljs#L29-L48) with [the electron one](https://github.com/atom/electron/blob/1bc49487add140f22ea4e454dcb0050e08679b4b/docs/api/app.md#appresolveproxyurl-callback) seems straightforward. What is currently unknown is how to verify this works. For anyone who is behind a proxy, please provide a means for the core team to reproduce being behind a proxy so we can implement and QA this

EarthCitizen:
For anyone looking for temporary solution:

The issue I found was that it was not able to connect through NTLM authenticated proxies. The workaround for me was to install and use Cntlm. It is a local proxy that connects to your network's proxy on your behalf. So, a proxy to your proxy. You give it the credentials to your network's proxy then just point LightTable at localhost.

wodin:
I think the only way to reproduce being behind a proxy would be to be behind a proxy :)
Maybe running a simple proxy on localhost would be enough?
e.g. something like this:
http://stackoverflow.com/a/7271863/495319
or maybe this:
https://github.com/nodejitsu/node-http-proxy

EarthCitizen:
I think specifically it was failing behind NTLM proxies, correct?

On Sunday, October 25, 2015, Michael Wood <notifications@github.com> wrote:

> I think the only way to reproduce being behind a proxy would be to be
> behind a proxy :)
> Maybe running a simple proxy on localhost would be enough?
> e.g. something like this:
> http://stackoverflow.com/a/7271863/495319
> or maybe this:
> https://github.com/nodejitsu/node-http-proxy
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/LightTable/LightTable/issues/1984#issuecomment-150900177>
> .
>


-- 
"Simplicity is the ultimate sophistication."
 --  Leonardo da Vinci

kenny-evitt:
[Squid](http://www.squid-cache.org/) might be a good candidate for something that could be setup so that a contributor could reproduce the issue(s).

[...]

PR #2208, opened 2016-05-16
Update Node package *request* to latest version.

kenny-evitt:
@cldwalker @rundis Several comments in #1984 claimed that upgrading the *request* Node package fixed issues related to proxy support, for downloading new versions of LT or its plugins.

Unfortunately, [there are a lot of changes that have been made to the package](https://github.com/request/request/blob/master/CHANGELOG.md) and, somewhat annoyingly, the latest version includes *65* other package dependencies. Should I have not included the ancillary dependencies? I'm guessing I should not.

The version we're currently using (2.14.0) doesn't have a corresponding tag or release in [the package's GitHub repo](https://github.com/request/request); the closest version with either (both actually) is 2.17.0.

We're only using this package (directly) in [one line](https://github.com/LightTable/LightTable/blob/53a90bee4c2c6758e4b5dffdffbc32256a05e00d/src/lt/objs/deploy.cljs#L81).


kenny-evitt:
I removed the 'extra' Node packages populated after running `npm install ...`.

kenny-evitt:
Running the function to download a specific release archive file works with the upgraded package. Updating the plugin metadata also seemed to work tho I just used the regular command to test that.

cldwalker:
@kenny-evitt Thanks for looking into #1984. Before we can merge this, we'll need to be able to QA this as we can't consider something fixed if we can't verify it ourselves. As for the extra dependencies, that does feel excessive. I don't think removing deps by hand is the solution as our `npm install` needs to be reproducible across machines. Since we just depend on it for a basic GET, we may be able to use a lighter package like https://github.com/bitinn/node-fetch or we could consider using the standard [http lib that comes with node](https://nodejs.org/api/http.html)

kenny-evitt:
@cldwalker I think requiring that we be able to QA this ourselves is excessive – in this case at least. Practically, requiring that we be able to reproduce the problem is almost certainly equivalent to us never fixing it. I know I'm not really motivated to fix it, especially if I have to setup a proxy somehow first.

I was motivated to submit this PR because the changes were so (apparently) minimal. And because someone already stated that the changes fixed their instance of the problem.

It also seems really unlikely that anyone is going to deliver detailed steps by which we can setup a proxy for ourselves, in a way that we can reproduce the original problem(s).

I think, in light of our limited resources, we can accept changes that (a) fix other's problems according to those same others; and (b) don't introduce any other problems.

The changes for this particular PR are so minimal, and affect such a small number of features, that I'm confident that they, *at worst*, fix at least *some* of the reported proxy problems (and don't break any existing features).

I think it's perfectly okay to treat this like any other 'environment' problem we can't reproduce. If someone complains about weird graphics problems, as we've seen before, and then later claims that upgrading one of our dependencies (e.g. Electron) fixes the problem, and then we confirm that upgrading the dependency doesn't break anything else, then it seems fine to merge changes to fix problems we can't ourselves directly QA.

Another alternative would be to merge these changes into a separate 'possible-proxy-fix' branch that interested users could build themselves. And, in fact, anyone like that can do that now.

So ... meh

As for the npm dependencies, I 'removed' some because it seemed like they were 'extraneous'. I tried running `npm install` *before* bumping the *request* package and just doing that 'added' three dependencies, i.e. resulted in three new sub-directories directly under *deploy/core/node_modules*, i.e. I was seeing 15 sub-directories under *node_modules* even tho there's only 12 in source. Do we need to fix that?

The build script doesn't actually run `npm install`; not under *deploy/core/* anyways. But none of the 'extra' packages seem to affect the build so I'm not sure why they're being added directly under *deploy/core/node_modules*.

I'm also fine replacing *request* with another library, but I think the next step should be either (a) reproducing the proxy problem among the core team; or (b) deciding we don't need to reproduce the problem ourselves.

cldwalker:
> I think it's perfectly okay to treat this like any other 'environment' problem we can't reproduce. If someone complains about weird graphics problems, as we've seen before, and then later claims that upgrading one of our dependencies (e.g. Electron) fixes the problem, and then we confirm that upgrading the dependency doesn't break anything else, then it seems fine to merge changes to fix problems we can't ourselves directly QA.

I don't think environment problems that we can't reproduce are in the same category as features we can't QA at all. The former are problems that we often don't have the hardware to do and are limited to very few users. Proxying is a feature we've known about since switching from nw.js and have tracked across multiple issues. It also effects many more users across all environments. If we're going to start supporting a feature, we need to have a bare minimum level of quality. Not being able to reproduce, understand or develop a feature falls below that bare minimum. By lowering that bar we create more work for ourselves in the future.

> As for the npm dependencies, I 'removed' some because it seemed like they were 'extraneous'. I tried running npm install before bumping the request package and just doing that 'added' three dependencies, i.e. resulted in three new sub-directories directly under deploy/core/node_modules, i.e. I was seeing 15 sub-directories under node_modules even tho there's only 12 in source. Do we need to fix that?

If you're using the node+npm versions in for-committers.md then it's worth looking into and understanding if it's an issue. To make our build more reproducible we should probably use npm-shrinkwrap.json alongside our package.json

kenny-evitt:
@cldwalker I'm fine with what you've written but I think this is a special case because we can, probably, fix this just by upgrading one of our existing dependencies.

If we were implementing the proxy support ourself then we should definitely be able to QA it ourselves too.

I think it's essentially riskless to bump the dependency version now and defer the ancillary issues indefinitely. Otherwise, realistically, this is going to take a lot longer to even potentially fix.

cldwalker:
Closing since this had extra files and now has conflicts. Can reopen once we decide to support and QA this feature

Issue #2218, opened 2016-06-03
Right click option to move file


HopefulLlama:
I can't find any option to move a file in my workspace.

kenny-evitt:
Please add [the info we ask for when reporting bugs in the contributing doc](https://github.com/LightTable/LightTable/blob/master/CONTRIBUTING.md#bugs).

I'll re-open when you've done that.

HopefulLlama:
Steps to reproduce: None - there is no option for moving files at all. What I would have expected is any of the following: 

- Right click on a file -> Move file option -> Move file dialogue
- Open command pane -> "Move File" -> Move file dialogue

Versioning: 
OS: Any
LightTable: Any
Plugins: None

Configuration: Default

Error Messages: None; I didn't do anything.

kenny-evitt:
@HopefulLlama thanks!

[Others have noted the lack of this "pretty basic feature that's missing".](https://groups.google.com/forum/#!topic/light-table-discussion/2fdVHOZQ9_U)

Are you interested in working on implementing this?

sbauer322:
@kenny-evitt would it be possible for you to outline steps/areas for this enhancement?

I should have some time to attempt this, but I am relatively new to LightTable's source and structure. Thanks!

kenny-evitt:
@sbauer322 I'm nearly as unfamiliar with the relevant code as you, but I'll gladly help out. I'll take a look at the code soonish. Ping me again in about a week if I don't follow-up with you before then.

Mouvedia:
If you take a stab at this try to implement the drag and drop in the workspace as well.

kenny-evitt:
@HopefulLlama Someone – @sbauer322 – submitted PR #2234 with changes that provide a move feature via the existing "Rename" item in the workspace context menu for files and directories. It'd be great if you tested those changes and let us know your thoughts. You'll need to clone the LT Git repo, merge the changes in the PR, and then build the source code to create a LT executable you can test. Let me know if you're interested in doing that and you need any help following the steps I outlined.

sbauer322:
@cldwalker. should this issue be closed, given that PR #2234 has been closed in favor of making it a plugin? Aside from drag and drop that @Mouvedia mentioned, the other options @HopefulLlama listed all would require essentially the same approach as that PR...

cldwalker:
@sbauer322 Worth leaving this open to get feedback once your plugin is available

PR #2234, opened 2016-07-30
Expand sidebar's rename functionality

sbauer322:
The Pull Request is for issue #2218. Rather than add a new, somewhat redundant, move function to the context menu in the sidebar, it was decided to expand the rename functionality.

It should be possible to now edit the file/folder name as well as the location of the file/folder within the filesystem. Below is an example of the UI at the start of the PR:

pr screenshot from 2016-07-30 00-37-03 cropped

This is currently a Work in Progress - looking for feedback! :smile:

### DIFF
--- a/src/lt/objs/sidebar/workspace.cljs
+++ b/src/lt/objs/sidebar/workspace.cljs
@@ -337,9 +337,6 @@
                                            (object/raise origin :rename.cancel)
                                            (object/raise origin :start-rename!))}]}))
 
-;; TODO: When moving file or folder into root folder (e.g., LightTable/untitled.txt)
-;; the file or folder does not appear until the folder is manually refreshed - all other deeper folders appear fine however
-
 (defn rename-file
   [origin path neue]
   ;; When moving a file, current-ws likes to be told about the change before
@@ -353,8 +350,9 @@
       (object/raise workspace/current-ws :watched.rename path neue))
     (files/move! path neue)
     (object/merge! origin {:path neue})
-    (remove-child (find-by-path old-parent-path) origin)
-    (add-child (find-by-path new-parent-path) origin)))
+    (when-not (or folder-root? file-root?)
+      (remove-child (find-by-path old-parent-path) origin)
+      (add-child (find-by-path new-parent-path) origin))))
 
 (defn rename-folder
   [origin path neue]
@@ -364,8 +362,9 @@
         new-parent-path  (files/parent neue)]
     (object/merge! origin {:path neue :realized? false})
     (files/move! path neue)
-    (remove-child (find-by-path old-parent-path) origin)
-    (add-child (find-by-path new-parent-path) origin)
+    (when-not (or folder-root? file-root?)
+      (remove-child (find-by-path old-parent-path) origin)
+      (add-child (find-by-path new-parent-path) origin))
     (let [docs (get-in @document/manager [:files])
           old-path (string/join [path files/separator])
           affected (filter (fn [x] (.startsWith x old-path)) (keys docs))]
--- a/src/lt/objs/sidebar/workspace.cljs
+++ b/src/lt/objs/sidebar/workspace.cljs
@@ -5,6 +5,7 @@
             [lt.objs.context :as ctx]
             [lt.objs.files :as files]
             [lt.objs.workspace :as workspace]
+            [lt.objs.topbar :as topbar]
             [lt.objs.opener :as opener]
             [lt.objs.popup :as popup]
             [lt.objs.sidebar :as sidebar]
@@ -29,6 +30,11 @@
 (defn root-file [path]
   (-> (object/create ::workspace.file path)
       (object/add-tags [:workspace.file.root])))
+
+(defn add-child [parent child]
+  (if (object/has-tag? child :workspace.file)
+    (object/update! parent [:files] (fn [cur] (vec (concat #{child} cur))))
+    (object/update! parent [:folders] (fn [cur] (vec (concat #{child} cur))))))
 
 (defn remove-child [p child]
   (if (object/has-tag? child :workspace.file)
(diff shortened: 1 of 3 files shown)


sbauer322:
This is a Work In Progress - there are still a couple of known bugs - the current purpose is to get eyes on the UI of the left sidebar's rename function.

@kenny-evitt if you could take a look at the UI to make sure it looks/feels appropriate that would be great!

src/lt/objs/sidebar/workspace.cljs
```diff
@@ -284,11 +285,32 @@
...
+           (object/raise this :rename.focus))
+  :blur (fn []
+          (object/raise this :rename.blur)))
+
+(object/object* ::rename-dialog
+                :tags #{:tree.rename}
+                :max-height 130
+                :path ""
+                :renaming? true
+                :behaviors #{::rename-focus ::rename-blur ::rename-submit ::rename-cancel ::rename-close! ::rename-file ::rename-folder}
```
sbauer322: @kenny-evitt, silly question, but is there a way to not specify the behaviours for the object and simply raise whatever behaviour? 

My question stems from the ::workspace.folder and ::workspace.file objects as they don't seen to have behaviours specified and work with various other behaviours... I could just be misunderstanding (probably am), so feel free to tell me otherwise. 

This is related to the PR as the previous rename functionality had two behaviours, rename-file rename-folder, both with the same trigger. Specifying the behaviours in the rename-dialog object is causing both rename-file and rename-folder to trigger as a result, which is not currently desired.

sbauer322:
I'm not sure if there is an interest, but rename seems like it might be something worthwhile to factor out from sidebar/workspace.cljs into its own file. Should that be a separate issue to open?

Mouvedia:
Just wanna say that you should take into account the discoverability. Windows users don't make the connection between rename and move as linux users do—due to the `mv` command.

First thing first have something that works.

> based on feedback

Id recommend the [human interface guidelines](https://elementary.io/docs/human-interface-guidelines) of Elementary OS. Of course you have to take into account the peculiar user base of LT. There's a big emphasis on the keyboard; in that regard:
- when esc is pressed you have to focus the last panel
- esc and outside click should have the same behaviour (close)
- keybinding/behavior for toggling (with auto text selection if present)
- the first time the text field appears for the current file it should highlight the filename excluding the extension (so that you can rename it directly)
- if closed the text should be saved (but not applied) so that when you reopen it is still there

Hope that helps.

Mouvedia:
> Create folders to an arbitrary depth when the user renames the path and it includes directories that do not currently exist.

I think a drop down—using auto complete—of the existing folders at that depth would make more sense. The user doesn't want to create folders he wants to either rename the file or move it to an existing folder. So the minimum would be to disable the rename if there's an inexistant folder (and highlight the culprit red). The harder solution would be to restrict the selection of the folders to the auto-complete suggestions.

sbauer322:
While I do agree these would be nice things, I do not feel they would be within the scope of this issue or appropriate to a essentially a first iteration.

That said, I intend to open additional issues recording those improvements after the PR is accepted so they can be addressed in the future.

Mouvedia:
So you intend to recursively mkdir?
That will probably be error prone in particular on case-insensitive filesystems.
Maybe add a visual clue for non-existing folders or a notification after the creation of a new folder.

IMHO this is too much; KISS and just log an error if a folder doesn't exist in the path.

sbauer322:
You do have a good point about it likely being error prone. I have removed the ability to add arbitrary folders as you have suggested. 

It should be pushed here sometime in the next few days.

sbauer322:
@kenny-evitt regarding the issue mentioned in Gitter:

When renaming a file or folder at the root folder level (not sure if that is the correct phrase, but I mean the highest point in a workspace) the file or folder disappears until the root folder is manually refreshed. Despite my best efforts, I cannot figure out why it is occurring only at this level - all other folder depths work properly...

As far as I can tell, renaming files and folders all work properly except for this level of folders. Would you have any ideas?

Here are some screenshots illustrating the problem:

Initial state - README.md is displayed in the sidebar properly

root-folder-does-not-update-initial

After renaming README.md to READMEs.md the file vanishes from the sidebar:

root-folder-does-not-update-after-rename

After manually refreshing the workspace folder READMEs.md will appear again:

root-folder-after-manual-refresh

This should be the last thing left to address before this pull request is ready (aside from future feedback)

src/lt/objs/sidebar/workspace.cljs
```diff
@@ -284,61 +291,132 @@
...
+        new-parent-path  (files/parent neue)]
+    (if (or folder-root? file-root?)
+      (object/raise workspace/current-ws :rename! path neue)
+      (object/raise workspace/current-ws :watched.rename path neue))
+    (files/move! path neue)
+    (object/merge! origin {:path neue})
+    (remove-child (find-by-path old-parent-path) origin)
+    (add-child (find-by-path new-parent-path) origin)))
+
+(defn rename-folder
```
sbauer322: Right around here would be the relevant rename-folder code, in case you wanted to check it regarding the above issue, @kenny-evitt

kenny-evitt:
The new file UI is interesting but I wasn't expecting it to look different:

image

I confirmed that I can move a file by renaming it – so that's good!

I'm going to stew on this for a while before I decide how I feel about it. Right now I'm conflicted. For those wanting to be able to their mouse and right-click, let alone drag-and-drop, this doesn't feel satisfying. And for those that wouldn't mind moving files by 'renaming' them, I'd expect they're already satisfied by just doing that thru a separate command line terminal program.

@rundis @cldwalker Your thoughts are most welcome.

sbauer322:
Yes, I agree those that want to use a mouse would not be satisfied. As mentioned at the start (via Gitter), I was using Atom as inspiration for this - they have something similar to what was implemented as well as the drag and drop functionality.

If the UI side is what is causing the conflict then it can be changed. I don't feel strongly about it... I would just need some idea of what it should look like instead.

I'm happy to include drag and drop as well but figured it should be a separate pull request - there were a lot of changes for a first time PR and I did not want to go overboard only to find out I was missing some critical thing.

Mouvedia:
Well normally if you want to move a file, you would expect a dialog box where you can navigate to select the destination folder. Of course all the files are hidden—only the folders are selectable.

sbauer322:
Totally agree and that was my first attempt! However, all of LT's dialog boxes (e.g., save, open) come from electron and I did not discover one for moving files.

Mouvedia:
The process would look like that:

1. select the folder using the file input
2. move the file(s) to the path returned

kenny-evitt:
@sbauer322 I forgot about the Atom inspiration. That does lend support for it not being completely foreign to everyone. How do you feel about hiding this behind a feature toggle, i.e. a setting that needs to be explicit set to enable this new UI feature?

Here's the Atom screenshot Scott shared in Gitter earlier (for anyone interested):

image

Can you share a full-window screenshot with the rename/move file UI from Atom too? I'm just curious what the visual difference is with the UI you've implemented.

Mouvedia:
> Enter the new path for the file.

Obviously shouldn't contain the file name as shown in the screenshot.

sbauer322:
Here you go @kenny-evitt: 

atom-full-ui-rename

Atom obviously has a more polished rename UI, but getting this UI hammered out is part of this PR too! This PR's rename UI was done rather simply since it was not clear how it should appear... should be fairly straightforward to make changes.

sbauer322:
With regards to having an explicit option to set for this to be enabled, I feel such a thing would be against the spirit of the original issue - they simply want some way to easily move files around within the IDE. Once drag and drop for files and folders is implemented then I would be more comfortable relegating this to an explicit option.

kenny-evitt:
@sbauer322 I created [a new post in the Google Groups group](https://groups.google.com/forum/#!topic/light-table-discussion/EP8bL-NhyME) asking for help testing and soliciting feedback about this.

I kinda like how your version of the UI is docked to the edge of the workspace sidebar/pane versus the Atom version being centered in the editor tab(s).

sbauer322:
Any chance we can merge this, @kenny-evitt? The Google Group post hasn't garnered any suggestions, or negative feedback, so I guess we are in the clear?

We can always add updates later on.

kenny-evitt:
@sbauer322 GitHub claims these commits conflict with the *master* branch. Please rebase *master* on your PR topic branch and fix any conflicts.

@rundis Mind reviewing this when you get a chance? I'm mostly okay with this but I'd prefer it 'hidden' behind a feature toggle so as to not surprise existing users. I'll defer to your judgement tho.

sbauer322:
@rundis, any chance I could get your feedback for this? I'm happy to make adjustments as needed... though I still feel this should not be hidden behind a feature toggle, I will do so if necessary.

One thought might be to extract all rename functionality as a whole into its own plugin... essentially start the process of decomposing Light Table itself into a series of plugins (there would be a core still, obviously).

cldwalker:
> The Pull Request is for issue #2218. Rather than add a new, somewhat redundant, move function to the context menu in the sidebar, it was decided to expand the rename functionality.

Is there relevant discussion on how this was decided? Not seeing in issues which is where any issue related discussion and decision should be happening. I'm not leaning towards this approach

> One (possibly absurd) thought might be to extract all rename functionality as a whole into its own plugin... essentially start the process of decomposing Light Table itself into a series of plugins (there would be a core still, obviously).

To that point, feature enhancements like this should be avoided in core unless we have a large user demand for it (which we don't here). I'm happy to consider updating LightTable core so that it could support a topbar plugin. Thoughts on what would need in core to support such a plugin?

Mouvedia:
> To that point, feature enhancements like this should be avoided in core unless we have a large user demand for it (which we don't here).

1. this is not a feature enhancement this is a pretty basic feature addition
2. [there is a demand for it](https://groups.google.com/forum/#!topic/light-table-discussion/2fdVHOZQ9_U)
3. what is essentially mv internally should be part of the core

> Not seeing in issues which is where any issue related discussion and decision should be happening.

I didn't see you participate in that thread: https://groups.google.com/forum/#!topic/light-table-discussion/EP8bL-NhyME

sbauer322:
Have to say you are sending mixed signals by adding this to the milestone list, but preferring it moved into a plugin, which would entail closing this PR! :wink: 

I want to say the discussion was either in Gitter, or simply decided on my own... the pull request itself was my first significant work on Light Table (prior to becoming a Light Table maintainer) and it was a way to learn about the internals. Since this enhanced rename contains the functionality of the original rename, I don't see much benefit in keeping them both.

As for adding a topbar, if you would rather see this as plugin, it shouldn't be that much different than our existing bars... something of a mix between bottombar, and popup. I don't know enough about plugins (yet) to say what would need to be exposed by Light Table to do this in a plugin, however.

cldwalker:
> Regarding demand for it, not having a way to move files around in Light Table itself is somewhat embarrassing. I can't think of any modern editor, let alone one listed in the top editors on Github, that don't have some version of this functionality... not to mention we like to describe ourselves as a "next generation code editor".

Light Table has abilities to rename a specific file and folder so I imagine it's that specific desire to move files across folders that you're looking for. General calls to feature parity and embarrassment aren't convincing

Looking at this PR more, there doesn't seem to be any major hooks for LT to provide for this as a [3rd party plugin](https://github.com/LightTable/plugin-metadata). If I am missing something on that front, let's discuss.

Overall I'm glad you're passionate about this functionality and that it could help some users. I'd recommend first publishing this as a personal plugin and pointing users to it in #2234. Over time we can discuss why this functionality merits being in core more than other 3rd party plugins. Feel free to close or I can soon

Mouvedia:
> General calls to feature parity and embarrassment aren't convincing.

That's  an unusual and peculiar stance. If he was asking for a diff/compare tool to be added to the core I would understand your position though. I don't understand how you can make a case for plugin. We probably have vey different ideas of what's *minimal*.

sbauer322:
Sorry, didn't mean to come across as particularly passionate on this topic... In the earlier comment, perhaps frustrating would have been more appropriate than embarrassing.

In any case, yes, being able to move files and folders around is what I was I looking to get out of this PR. 

As for merit to include this in core, we already have similar functionality in the core, which is our current rename. Splitting the two - with one in core and the other a plugin - seems odd. It should be one way or the other. But all this goes back to if we wanted to pursue breaking up Light Table functionality, such as rename/move file and searching, into their own plugins.

Additionally, #2218 was labelled as an enhancement as opposed to being closed or added to our plugin wishlist. 

Beyond what was said above, I am not going to try to attempt anyone of the merit of this to be in core - it simply is not that big of an issue to squabble over. I'm fine with closing this and relegating it to a plugin someday.

cldwalker:
> Additionally, #2218 was labelled as an enhancement as opposed to being closed or added to our plugin wishlist.

@sbauer322 I set up these labels. An [enhancement label](https://github.com/LightTable/LightTable/wiki/For-Contributors/4a9c3835424cb28b61a2e9ca2e5585582a3be27e#issue-labels) isn't a reason or a substitution for a discussion. An enhancement is an "Interesting feature suggestion which isn't on the roadmap but welcome". If you're using labels in a different way, please define them somewhere

sbauer322:
I'm not sure what more I can say, but there was quite a lot of discussion, and it was okayed to implement in Light Table core and to expand the current rename functionality.

Some relevant links to the Gitter discussion, which is where we are directed to post questions and other transient information:

- [Proposing we adopt Atom's style of renaming for the PR](https://gitter.im/LightTable/LightTable?at=578eb6d82d04e058518f9361)
- [@kenny-evitt okaying the style](https://gitter.im/LightTable/LightTable?at=578ed5690720fd587aa393d8)
- [@kenny-evitt] okaying to use Gitter for my questions relating to the PR](https://gitter.im/LightTable/LightTable?at=578fbaa73d74e5a0167582a2)
- [Asking about rename vs move](https://gitter.im/LightTable/LightTable?at=57953c15b92e27447fb23f21)
- [Stating that I am going to merge the functionality of rename and move then call it rename](https://gitter.im/LightTable/LightTable?at=57954985959ee82b3e532450)
- [@kenny-evitt okaying all of this](https://gitter.im/LightTable/LightTable?at=579aaf3fac80b5ea3f1516d2)

cldwalker:
> I'm not sure what more I can say, but there was quite a lot of discussion, and it was okayed to implement in Light Table core and to expand the current rename functionality.
> 
> Some relevant links to the Gitter discussion, which is where we are directed to post questions and other transient information:

Our policy on [feature discussion](https://github.com/LightTable/LightTable/blob/master/CONTRIBUTING.md#feature-requests) is pretty explicit. Apologies if you were lead to think that didn't apply for this case. For future reference, gitter is fine for support and realtime discussion but ultimately LightTable work needs to be coordinated on github

> Regarding the label, I am aware of that definition and it is fine... my point was that it is misleading to label an issue as an enhancement or feature while expecting it to actually be a plugin. It only helps to cause confusion and we should be more clear about what issues we expect a plugin rather than a pull request.

The whole point of the label descriptions is to minimize this confusion. Naming alone doesn't provide any of the context on how the core team processes these labels. I've brought back the [descriptions on the wiki](https://github.com/LightTable/LightTable/wiki/For-Contributors#issue-labels). Please update the ones you've changed. If you have better names for the `enhancement` description I'm open to them

sbauer322:
The label names have been restored.

Issue #2237, opened 2016-08-05
Replace wrench with shelljs


sbauer322:
According to wrench's [github page](https://github.com/ryanmcgrath/wrench-js), it has been deprecated and the recommended replacement is [fs-extra](https://github.com/jprichardson/node-fs-extra). Since we don't need everything that fs-extra offers, replace wrench with [shelljs](https://github.com/shelljs/shelljs).

sbauer322:
Updated to reflect that we are not pursuing fs-extra and instead going to use shelljs

Issue #2129, opened 2016-01-21
Save-file-as and workspace "Duplicate" menu command don't update the workspace bar


kenny-evitt:
This can be reproduced on both Windows (Windows 7) [note for myself – *T*], Mac OS X, and Linux (Ubuntu).

Steps to reproduce for save-file-as:

 1. Open a tab for an existing file.
 2. Use any of the following:
  a. Open the *File* menu and pick *Save file as..*.
  b. Open the command bar and pick the command *File: Save file as..*.
  c. Enter the default keybinding (assuming it's still bound) <kbd>Ctrl</kbd>+<kbd>Shift</kbd>+<kbd>s</kbd>.
 3. In the OS file dialog window, pick a directory for which files are listed and visible in the workspace bar in Light Table, enter a new name for the file, and save the new file with the new name.
 4. The new file won't be automatically listed in the workspace bar; it should be.

kenny-evitt:
This was discovered while QA-ing the 0.8.1 release candidate build; see #2126.

cldwalker:
I suspect https://github.com/LightTable/LightTable/pull/2105 but will require more investigation

sbauer322:
This was rediscovered during work for PR #2234, but can still be reproduced on 0.8.1 via the context menu's Duplicate. Rather than create another extremely similar issue, the relevant information is below with screenshots for reproducing the issue with Duplicate from the context menu. 

On LT 0.8.1 this happens at all folder levels.

Here are some screenshots illustrating the problem (and workaround):

Initial state - 'README.md' is displayed in the sidebar

![sidebar-pre-duplicate](https://cloud.githubusercontent.com/assets/2945386/17641254/ed58e906-60e5-11e6-84ee-50135cbd2452.png)

Open context menu on 'README.md' and select Duplicate

![sidebar-duplicate](https://cloud.githubusercontent.com/assets/2945386/17641256/f33ef3e2-60e5-11e6-8ccd-9616f91e87be.png)

Nothing appears

![sidebar-post-duplicate](https://cloud.githubusercontent.com/assets/2945386/17641260/ff5b7bb4-60e5-11e6-9037-bc9de823d734.png)

Open context menu on root folder and select Refresh

![sidebar-refresh-folder](https://cloud.githubusercontent.com/assets/2945386/17641263/0ae09802-60e6-11e6-8ada-02b37b253714.png)

Duplicate file 'README copy.md' is now visible 

[...]

PR #2238, opened 2016-08-13
Replace wrench with shelljs

sbauer322:
In this pull request for issue #2237 wrench is swapped out for fs-extra due to the deprecation of wrench. Luckily wrench was only used in two places making this a relatively minor pull request.

There has been no change in functionality with regards to delete and copy within files.cljs as a result.

If desired, we can expand the scope of this PR to update all places we use fs to use fs-extra instead. fs-extra contains fs within it...


sbauer322:
Please verify everything is hunky dory with regards to npm. After updating package.json, the commands used were:

npm install fs-extra@0.30.0
npm uninstall wrench

And the results committed

kenny-evitt:
@sbauer322 Thanks for this PR! Sorry for taking so long to get around to reviewing it.

But do we even need *fs-extra*? I don't think we do. The *shell* npm package `rm` method seems to handle recursive deletes just fine. The existing upgrade code has been using it for three years. Would you mind testing whether we can simply use the *shell* package methods instead?

sbauer322:
Will do! I had chosen fs-extra simply because that is what wrench
suggested.

On Aug 14, 2016 12:33, "kenny-evitt" <notifications@github.com> wrote:

@sbauer322 <https://github.com/sbauer322> Thanks for this PR! Sorry for
taking so long to get around to reviewing it.

But do we even need *fs-extra*? I don't think we do. The *shell* npm
package rm method seems to handle recursive deletes just fine. The existing
upgrade code has been using it for three years. Would you mind testing
whether we can simply use the *shell* package methods instead?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/LightTable/LightTable/pull/2238#issuecomment-239682869>,
or mute the thread

rundis:
If we can remove npm dependencies I'm all for that. Funnily I happen to use wrench from a couple of my plugins, but that's obviously of no concern to LT. But it should come as a warning in the release notes. Hopefully there aren't many other plugins that use it.

kenny-evitt:
@sbauer322 Thanks!

@rundis You and other plugin authors can pull in *wrench* (or any other Node.js package) independently of LT tho – right? Is the potential issue just that removing it from LT core might break plugins that are using it without including it in their own projects themselves?

rundis:
The latter.
I haven't found anything in the dev docs stating that you shouldn't use the node modules shipping with LT. Maybe it should mention that if you do it's at your own peril, because it's not part of the official "api" for plugins ?

kenny-evitt:
@rundis We should mention that those packages aren't part of the 'official API' – now I'm trying to figure out where best to do that! Any ideas?

sbauer322:
@kenny-evitt my preference would be stating this somewhere within for-committers.md, the wiki, and package.json itself as a comment. It is possible to miss if it is only within a single supporting document.

sbauer322:
I was able to switch `files.cljs` to use [`shelljs`](https://github.com/shelljs/shelljs) instead of `wrench`. Do note, there is a different library(?) named [`shell`](https://github.com/electron/electron/blob/master/docs/api/shell.md) already being used within `files.cljs`.

Sorry for the list, @kenny-evitt, but there are a couple questions now:

- Should we rename what is currently known as shell to something such as `electron-shell` across the project? It can be easily confused with shelljs if you are not aware of both of them... I know I wasn't! :smile: 
- Do you want wrench removed as a dependency within LT or leave it? Alternatively, we could deprecate it for future releases and aim to remove it entirely for 0.9 or 1.0?

rundis:
Tx for your contributions @sbauer322. Much appreciated.
I'd say we go for the following. 
- remove wrench completely
- either rename use of shell to electron-shell as you suggested, or  rename var usagee of shelljs to be named shell-js. Prefer the former.

sbauer322:
Should be all set now @rundis, @kenny-evitt! 

Wrench has been removed, shell has been renamed to electron-shell and shelljs is now shell

kenny-evitt:
@sbauer322 Please rebase the upstream *master* onto this branch.

kenny-evitt:
@sbauer322 @rundis I just tested all of the UI features that use any of the affected vars and everything looks good (or as good as it was before these changes). The only issue I noticed was #2129 but Scott already commented that it's still an issue even with the other related changes.

@sbauer322 I'll merge this after you've rebased the upstream *master* on top of this branch.

sbauer322:
@kenny-evitt: I believe I have done what you asked, but full disclosure - I've never used rebase before. 

The commands used were:

```
git checkout replace-wrench-with-fs-extra
git rebase LightTable/master
```

Could you confirm I did not do something undesired prior? If all is well then merge away!

kenny-evitt:
@sbauer322 I'm not sure what happened. GitHub doesn't show any changes for the 'merge' commit. I think it's okay. Would you squash the three commits as one commit?

sbauer322:
I think we are clear to merge now, @kenny-evitt!

src/lt/objs/files.cljs
```diff
@@ -231,13 +233,13 @@
...
 (defn trash! [path]
-  (.moveItemTotrash shell path))
+  (.moveItemTotrash electron-shell path))
 
 (defn delete!
   "Delete file or directory"
   [path]
   (if (dir? path)
-    (.rmdirSyncRecursive wrench path)
+    (.rm shell "-rf" path)
```
cldwalker: @sbauer322 Using the force option here is a little concerning. Do you have a specific use case for it? Given that [this option ignores write permissions](https://github.com/shelljs/shelljs/blob/35cef9a4ae2b7321bd8cc341a7528ef015bca304/src/rm.js#L125-L127), any bugs (accidental args) to this fn could lead to harmful deletions for users, especially those running as root

sbauer322: Nice catch! You are correct, it does not appear force is needed based on some quick testing I just did. I have opened PR #2284 to address this.

PR #2239, opened 2016-08-14
Add note about removing a Node package.

kenny-evitt:
[Magnus pointed out that removing a Node.js package from the LT core source might affect plugins.](https://github.com/LightTable/LightTable/pull/2238#issuecomment-239687875) I (stupidly) didn't think of that.

I think I should also add a checklist item to review changes for a release and check for any that remove Node.js packages so that users can be notified of the change as described in the note I added already for this PR.

 - [x] Do the thing I just described in the paragraph above.

### DIFF
--- a/doc/for-committers.md
+++ b/doc/for-committers.md
@@ -15,6 +15,8 @@ Current ClojureScript version and libraries we use are in [project.clj](https://
 
 Node package installs last done with node.js v2.5.0 and npm v2.13.2.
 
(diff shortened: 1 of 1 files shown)


kenny-evitt:
@rundis Inspired by your comment on #2238 pertaining to my asking to remove a Node.js package from LT core.

sbauer322:
This seems good to me too. I added a question in the Dev FAQ indicating to not use LightTable's internal node modules for plugins.

rundis:
Cool! Tx.

Issue #2254, opened 2016-09-06
Add documentation for whole file evaluation


sbauer322:
Stemming from discussions on https://github.com/LightTable/LightTable/issues/2170 it would be beneficial to add documentation for whole file evaluation, especially when eval'ing LightTable itself, or at least move the information gained from the discussion into a document in the project.

Issue #2170, opened 2016-03-25
Can’t eval editor.cljs in LT UI


martinchooooooo:
#### To reproduce
- Have LT source in your workspace
- open `src/lt/objs/editor.cljs`
- Eval whole file (for me it’s pmeta-shift-enter)
- Get error `#< Error: This document is already in use >`
- Now when trying to use LT, nothing really works. For example try using find in the editor and you’ll get errors in the console.

#### Environment:
- LT 0.8.1, built from master (a8a4f2e)
- OSX 10.11.1
- [user.behaviors and user.keymap](https://gist.github.com/martinchooooooo/9ecfbf03f66a74a2916d)

#### In trying to track it down..
(In the process went down the rabbit-hole trying to figure out how the whole BOT thing works with a bunch of print statements, which I’ve been wanting to do for a while so it was a bit of fun :smile:)
- The error is thrown inside of CodeMirror in [function attachDoc](https://github.com/codemirror/CodeMirror/blob/36821970d225d430fb6d6dc6612d259c75d28d4d/lib/codemirror.js#L7754), which we call through `swapDoc` in `editor.cljs` ([source](https://github.com/LightTable/LightTable/blob/f07d4455b1c46fb6bf0d6a5eb7799d71d492adb5/src/lt/objs/editor.cljs#L100))
- For those not particularly familiar with LT code base, like I was, the call is being made upon creating the editor object in the [:init function](https://github.com/LightTable/LightTable/blob/f07d4455b1c46fb6bf0d6a5eb7799d71d492adb5/src/lt/objs/editor.cljs#L464), which calls (make) and there you have the the swapDoc call.
- The editor object is created by `editor/pool.cljs` [here](https://github.com/LightTable/LightTable/blob/f07d4455b1c46fb6bf0d6a5eb7799d71d492adb5/src/lt/objs/editor/pool.cljs#L183)  and `opener.cljs` [here (i think?)](https://github.com/LightTable/LightTable/blob/f07d4455b1c46fb6bf0d6a5eb7799d71d492adb5/src/lt/objs/opener.cljs#L149), so perhaps this is where we have to do some kind of check to see if the editor we’re trying to create already exists?
- Not quite sure what what CodeMirror is checking for in performing the test, but it doesn’t seem to like that the editor object we’re trying to create in LT when evaluating editor.cljs
- If i try to eval some other file, such as `pool.cljs`, the eval is fine with no problem.
[...]

PR #2257, opened 2016-09-10
Add info about whole-file eval for Light Table source files to 'committers' doc.

kenny-evitt:
See #2254 and #2170, which inspired the former.


rundis:
Looked good to me. Tx !

Issue #2203, opened 2016-05-11
unnable to get lighttable docstrings


carocad:
This happens on my LT v0.8.1 Ubuntu 14.04

When programming a LT plug-in, it is impossible to get the documentation of lighttable's internal vars. For example: pressing (Ctrl+D) 'toggle documentation at cursor' on 'defui' shows nothing, the same happens for any other LT internal function or macro. Searching the language docs doesn't work either; 'Found 0 results'.

I guess that this is due to LT internals being stored in LT itself and not on the nREPL but this is just a guess.

To reproduce this simply, open a plugin source code and try to get documentation on any Lighttable var. A widget with the current namespace will be _incorrectly_ displayed.

carocad:
closing since the second PR was already merged

PR #2276, opened 2016-09-29
Add optional docstring to defui, behavior, and object - closes #2206

### LINKED ISSUE #2206: optional docstring for lt macros
hey guys,
to be honest I have never written a macro so I don't know if this is actually possible or not, but anyway here it goes. I was having troubles understanding some of the uses of some UIs and also some behaviors, the problem being that they don't have any documentation which is not a surprise.

The real problem though is that (from my limited understanding of macros) it is also not possible to (currently) add docstrings to them since the macro was never intended for it. This is true for both `defui` and `behavior` although I am not sure how plausible it would be with `behavior` since the name gets munged.

```
;; currently
(defui hello-panel []
  [:h1 "Hello from the playground"])

;; expectation
(defui hello-panel
"creates a hello panel as an h1 tag"
 []
  [:h1 "Hello from the playground"])
```

I think that adding the option of docstring for the BOT elements would make understanding a plugin code much easier as currently we have to go with the best guess based on name, triggers and similars.

sbauer322:
This pull request adds an optional docstring to `defui`, `behavior`, and `object` for issue #2206. It is viewable by the in-line doc command (Ctrl-D). The key to use is `:doc` for behaviors and objects. defui is similar to defn.

One caveat is that the docstrings for behaviors and objects will not appear in Codox. This should not be an issue as behaviors and objects do not currently appear in Codox either... Behaviors likely won't be visible in Codox unless their internal functions are made public, which is probably a bad idea. Objects have a similar issue.

Here are samples:

object-docstring

behavior-docstring

Do note, this issue does not deal with tags or triggers... so using Ctrl-D on trigger `:close!` rather than the behavior `::close!` will result in no docs found. I do not feel that adding logic to get behavior or object docstrings through tags or triggers will be feasible... but I am not certain - something to investigate in another issue!

### DIFF
--- a/src/lt/plugins/doc.cljs
+++ b/src/lt/plugins/doc.cljs
@@ -76,12 +76,46 @@
               (not= (:doc doc) "nil"))
      [:pre (:doc doc)])])
 
+(defn- retrieve-behavior
+  "Helper method for behavior `editor.doc.show!` to determine if the given `ns` and `name`
+  match existing behaviors.
+
+  Returns found behavior or `nil`."
+  [ns name]
+  (@object/behaviors (keyword (str ns "/" (subs name 2)))))
+
+(defn- retrieve-object-def
+  "Helper method for behavior `editor.doc.show!` to determine if the given `ns` and `name`
+  match existing object defs. Not recommended to print whole object def... use destructuring.
+
+  Returns found object def or `nil`."
+  [ns name]
+  (@object/object-defs (keyword (str ns "/" (subs name 2)))))
+
+(defn- retrieve-docstring
+  "Helper method for behavior `editor.doc.show!` that returns the docstring for a matching
+  object or behavior. If `:doc` is not found, then `:desc` is used. Otherwise `nil`."
+  [ns name]
+  (let [beh (retrieve-behavior ns name)
+        beh-doc (or (:doc beh) (:desc beh))
+        obj (retrieve-object-def ns name)
+        obj-doc (or (:doc obj) (:desc obj))]
+    (if (and (nil? beh-doc) (nil? obj-doc))
+      nil
+      (if (nil? beh-doc)
+        obj-doc
+        beh-doc))))
+
 (behavior ::editor.doc.show!
           :triggers #{:editor.doc.show!}
           :reaction (fn [editor doc]
                       (when (not= (:name doc) "")
-                        (inline-doc editor (doc-ui doc) {} (:loc doc)))
-                      ))
+                        ;; If :file and :doc are nil then this is likely a behavior or object.
+                        ;; Check if a match exists and splice the resulting :doc into the doc argument.
+                        (let [doc (if (and (nil? (:file doc)) (nil? (:doc doc)))
+                                    (merge doc {:doc (retrieve-docstring (:ns doc) (:name doc))})
+                                    doc)]
+                          (inline-doc editor (doc-ui doc) {} (:loc doc))))))
 
 (defui search-item [item]
   [:li
--- a/src/lt/macros.clj
+++ b/src/lt/macros.clj
@@ -12,7 +12,8 @@
   * :reaction (required) - Function to invoke when behavior is called.
                            First arg is object behavior is attached to
   * :triggers (required) - Set of keyword triggers that trigger behavior
-  * :desc - Brief description of behavior
+  * :desc - Brief description of behavior.
+  * :doc - Equivalent to a traditional function docstring.
   * :type - When set to :user, shows up in hints. Not enabled by default
   * :params - Vector of maps describing behavior args. Each map contains required :label key
               and optional keys of :type (:string, :number or :list), :items and :example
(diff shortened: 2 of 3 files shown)


sbauer322:
I feel adding a separate issue to make the optional docstring view-able by the in-editor docs, and potentially Codox, is the best course of action should not having object and behavior docstrings be a sticking point.

sbauer322:
@LightTable/committers I have finished altering this PR and it is ready for review.

src/lt/plugins/doc.cljs
```diff
@@ -76,12 +76,44 @@
...
+
+  Returns found object def or `nil`."
+  [ns name]
+  (@object/object-defs (keyword (str ns "/" (subs name 2)))))
+
+(defn- retrieve-docstring
+  "Helper method for behavior `editor.doc.show!` that returns the docstring for a matching
+  object or behavior, if any exist, or `nil`."
+  [ns name]
+  (let [beh-doc (:doc (retrieve-behavior ns name))
```
rundis: Applies to both:
- Would it make sense to fall back to using :desc if no :doc provided (people are lazy, and some doc is better than no doc, even though :desc is sometimes more of a end user facing thing)

src/lt/plugins/doc.cljs
```diff
@@ -76,12 +76,44 @@
...
+        obj-doc (:doc (retrieve-object-def ns name))]
+    (if (and (nil? beh-doc) (nil? obj-doc))
+      nil
+      (if (nil? beh-doc)
+        obj-doc
+        beh-doc))))
```
rundis: It would be cool to extract any `:label`  values from the `:params`  vector. That way in some instances you would be able to pass the args key to the inline doc thingie.

rundis: Oh an immediate benefit of a fall back to desc is that we quickly get access to what's already there. Behaviors and objects are one of  the "best" documented parts of LT (bar yours and @cldwalker s efforts to document namespaces and functions).

sbauer322: Having desc as a fall back should be straight forward enough and I will include it in this PR.

Would you be opposed to having the label values added in a separate issue and PR? It sounds useful, and will implement it, but I prefer to avoid adding a bunch of additional stuff beyond the original scope of the issue.

sbauer322: Yes, your line of thinking about adding the additional :doc key is correct. As :desc is user facing, and should be limited to something less than a full sentence, it made more sense to add a separate key that would function like a traditional docstring.

rundis: sure make a separate issue for the params/labels thing.

sbauer322: All set @rundis, this will happen with both objects and behaviors:

object-docstring-with-desc-fallback

sbauer322: Is this in the clear to merge, @rundis? I have opened #2281 for the :params and :label request.

rundis:
Ok So I gave it a test run locally, and had two issues

1. Docstring on defui, what am I missing ?
screen shot 2016-10-06 at 07 31 14

2. Using dot (`.`) in behavior name or object name doesn't seem to work too well
screen shot 2016-10-06 at 07 34 40

I'm guessing this might be the root cause: https://github.com/LightTable/Clojure/blob/master/src/lt/plugins/clojure.cljs#L610

sbauer322:
1. I am able to reproduce that error on 0.8.1 if I don't build and launch Light Table. In other words, it doesn't have the optional docstring work included. Maybe try rebuilding Light Table? My flow is:

```
lein cljsbuild once app
./script/build-app.sh
builds/lighttable-0.8.1-linux/LightTable
```

Otherwise, I'm not sure what the problem is... it works fine for me :grin: 

defui-docstring
2. I concur that the link you provided might be the root cause. It could also explain why I have never noticed any behaviors or objects named that way... usually they have a dash from what I see (coming from a non-plugin background at least). This happens in Light Table 0.8.1 also... so another issue found! :smile:

rundis:
I checked out master to a clean dir. Fetched the pull request and checked out the resulting pr branch. Then ran script/build.sh and finally launched by invoking `builds/lighttable-0.8.1-mac/light`. That to my mind should have been enough to get the latest and greatest in your pull request ?

sbauer322:
I would try to do the `lein cljs build once app` before the build.sh as there are changes to the Clojurescript files that need to be propagated.

rundis:
Sorry. Well yeah the behavior macro change deffo works, because the behavior thing works (showing the extra doc key if I add it).
- running the lein command wont help that's exactly what is being run inside build.sh

sbauer322:
D'oh! I've been thinking of  `build-app.sh` but referring to `build.sh` this whole time. Yes, you are correct that `build.sh` runs the lein command.

Seems we are in a pickle then where it works for me, but not you... are you testing out the functionality within the Light Table codebase, or in something else like a plugin?

rundis:
In the user plugin

sbauer322:
I was able to reproduce this in the User plugin, so at least there is that! As for the why - I am still looking into this.

rundis:
Maybe, but in this case I was trying to get the doc string for something I evaled in the plugin (ie plugin objects, behaviors and defui). I still find it weird that it worked for behaviors, but not for defui. I can't really spot what could possibly be wrong with the defui macro (but then again my macro-fu is minimal).

#2203 is a bummer yeah.

sbauer322:
That is a good point about how behaviors and objects created in the User plugin will work... It has been quite an amusing problem. 

As for why behaviors and objects work and defui does not, I suspect it stems from being able access the behavior and object docs via atoms in the object namespace... you simply have to know the keyword the behavior or object is under. Defui does not have something like this. 

A more arbitrary guess, based on an above comment, is that the inline docs command is not sending all the parameters. I was able to see this last night by setting the parameters themselves as the doc to return and comparing side by side the User plugin and Light Table core.

I'll give it another go tonight and maybe tomorrow, but I am not confident a solution will be found... after the weekend, I'd recommend saving this problem for #2203 as defui is explicitly mentioned in it.

sbauer322:
I will attempt it later, but one potential solution is to simply store defui stuff in an atom like we do with behaviors and objects in the object namespace.

sbauer322:
I've not had much luck in ascertaining what causes plugins to behave differently, @rundis. I'm out of ideas, but it almost feels like the connection is not using the Light Table build with the docstring changes.

In light of this, shall we defer searching for a solution to this particular problem and wait for #2203?

sbauer322:
@LightTable/committers any thoughts on how to proceed, or are we clear to merge? I feel getting plugins working with this will ultimately fall under #2203 as the problem at hand is not obvious and outside of this issue's scope.

rundis:
I´m ok with merging. Incrementally better and still useful

cldwalker:
> @LightTable/committers any thoughts on how to proceed, or are we clear to merge? I feel getting plugins working with this will ultimately fall under #2203 as the problem at hand is not obvious and outside of this issue's scope.

@sbauer322 If there are known issues with this feature please create an issue for it. #2203 doesn't capture the specific issue here spread across multiple comments

I'm a little disappointed this was merged without resolving underlying issues. While it's great to have contributions from new team members, we should avoid contributions that are incurring new issues right off the bat

sbauer322:
https://github.com/LightTable/LightTable/issues/2293 has been created to capture the situation where the defui docstring does not work in plugin development. 

I agree this came out underwhelming due to the defui docstring plugin incompatibility, but I am not aware of it incurring new issues. Could you be more specific?

This PR might have made other preexisting issues, like #2203 or dots in behavior names, more obvious and pressing, but they existed prior. Both are in my unofficial to-do-eventually pile, but I'd be happy to have a more senior team member provide contributions for them.

If it would make you more comfortable, feel free to revert the changes.

PR #2284, opened 2016-10-09
Remove unnecessary, and potentially dangerous, -f from delete! function.

sbauer322:
This PR removes the -f parameter from shell's rm command in files/delete!. As @cldwalker noticed, it is not needed and could be a potential source of trouble.


cldwalker:
Looks good. Thanks for the quick turnaround! :+1:

PR #2289, opened 2016-11-08
Plugin manager changes.

chris-etheridge:
* Change the `source` button to `website`.
* Clicking the title of a plugin will open the `source` page.

### DIFF
--- a/src/lt/objs/plugins.cljs
+++ b/src/lt/objs/plugins.cljs
@@ -494,7 +494,7 @@
    (search-input this)])
 
 (defui source-button [plugin]
-  [:span.source [:a {:href (:url plugin (:source plugin))} "source"]]
+  [:span.source [:a {:href (:url plugin (:source plugin))} "website"]]
   :click (fn [e]
            (dom/prevent e)
            (dom/stop-propagation e)
@@ -526,6 +526,13 @@
            (dom/prevent e)
            (dom/stop-propagation e)))
 
+(defui plugin-title [plugin]
+  [:h1 (:name plugin) [:span.version (:version plugin)]]
+  :click (fn [e]
+           (dom/prevent e)
+           (dom/stop-propagation e)
+           (platform/open-url (:url plugin (:source plugin)))))
+
 (defui server-plugin-ui [plugin]
   (let [info plugin
         ver (:version info)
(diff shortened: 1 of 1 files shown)


src/lt/objs/plugins.cljs
```diff
@@ -494,7 +494,7 @@
    (search-input this)])
 
 (defui source-button [plugin]
```
sbauer322: No need to do the below in this PR, but just taking the opportunity to open a potential conversation:

One question I would like to generally propose to anyone reading is if we should change the code from `source` to `website`... such as the defui's `source-button` to `website-button`. Minor detail, but I feel this could potentially break custom themes. Probably best to simply open an issue to address it in the future.

sbauer322:
Thanks for opening this! It looks good to me, but I am going to leave it open until November 12, 2016 to give others a chance to comment and/or provide feedback.

src/lt/objs/plugins.cljs
chris-etheridge: So I was actually going to bring this up, but didn't want to make the PR fix two fairly different things.

Whilst I was implementing, and opening, this PR, it became confusing when using `source` and `website` interchangeably. I think it would be a good idea to refactor the wording, so that it matches what the UI shows? I don't mind doing this either :)

chris-etheridge: Opened an issue [here](https://github.com/LightTable/LightTable/issues/2290).

sbauer322: That's perfectly understandable. :smile:

Since I'm not sure if it would break customized themes, I am leaning towards leaving it as another issue (looks like issue #2290) and PR that would be merged in the future with along other potentially breaking changes we have planned.

chris-etheridge: Cool! Well I'm happy to leave it open till other people have a say in it.

sbauer322:
@LightTable/maintainers any thoughts before I merge this?

cldwalker:
@sbauer322 This PR looks good. A linkable title is a nice improvement. Indifferent about button wording change. I do care more that such changes can lead us down refactoring and thus potential plugin breakages

sbauer322:
I apparently forgot to merge this, but it works great, @chris-etheridge - thank you!

I did notice one slight improvement to make it more obvious that the title is a link and opened https://github.com/LightTable/LightTable/issues/2294 to address this.

PR #2296, opened 2016-11-29
Update style guide to define expected :desc convention

sbauer322:
Simply documents an undocumented convention.


sbauer322:
I will be merging this on December 10, 2016 unless there is any feedback, @LightTable/maintainers

cldwalker:
Thanks for the update! :+1:

PR #2301, opened 2016-12-22
Add label field to the inline doc for behaviors and objects - fixes #2281

### LINKED ISSUE #2281: Add label values to inline doc for objects and behaviors
With PR #2276, behaviors and objects now have useful info in their inline documentation (Ctrl-D). Another useful piece of information are the `:label` values from the `:params` vector. That way, in some instances, you would be able to pass the args key to the inline doc.

Below, an example of `:label` and `:params` can be seen in use, but not displayed in the doc box:

object-docstring-with-desc-fallback

sbauer322:
This PR adds a parameter label field to the in-line doc for behaviors and objects for #2281. If there are no parameters with labels listed in the map then the line is omitted.

Below is an image showing a behavior with a parameter map and corresponding parameter labels:

### DIFF
--- a/src/lt/plugins/doc.cljs
+++ b/src/lt/plugins/doc.cljs
@@ -72,6 +72,9 @@
    (when (and (:args doc)
               (not= (:args doc) "nil"))
      [:h3 (:args doc)])
+   (when (and (:labels doc)
+              (not= (:labels doc) ""))
+     [:h3 (str "[" (:labels doc) "]")])
    (when (and (:doc doc)
               (not= (:doc doc) "nil"))
      [:pre (:doc doc)])])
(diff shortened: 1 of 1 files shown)


sbauer322:
If no one has any input, I am going to merge this on January 7th... but it would be nice to at least hear from @rundis.

rundis:
Looks good to me. 
Possible additions for another issue could be:
- Also display the `:type` attribute/key for params (when given)
- pair with arg-names from the behavior reaction function :-)

sbauer322:
https://github.com/LightTable/LightTable/issues/2312 and https://github.com/LightTable/LightTable/issues/2311 have been created for those possible additions.

PR #2318, opened 2017-01-31
Fix zoom keybindings - fixes #2317

### LINKED ISSUE #2317: Fix default zoom keybindings for numpad use
This is similar to issue #2181, but I am experiencing a few inconsistencies related to zoom's default keybindings:

- `pmeta`-`-` works fine.
- `pmeta`-`+` is not a default keybinding, but is actually bound to `pmeta`-`=` and `pmeta`-`Shift`-`+`... so it does not work with the numpad.
- `pmeta`-`0` is to reset zoom, but does not work with the numpad either.

We will want to add the appropriate bindings to allow for numpad use.

sbauer322:
Basically, Mousetrap had a bug preventing the numeric pad `0` from working properly... instead it was being registered as a backtick, `. Since the bug was fixed in a later commit of Mousetrap, I have updated it.

To better facilitate keeping our Mousetrap up to date I have forked the repo under the Light Table org until we can remove our changes to the library. These changes are now clearly indicated in the forked mousetrap.js (renamed to keyevents.js in LT).

Finally, the Mousetrap section of `for-committers.md` has been expanded.

I should add that I did nothing to address the keybinding conflict on `Ctrl`-`=` as I am not sure the best way to proceed, but instead added a keybinding for `Ctrl`-`Shift`-`=` as a workaround. This conflict seems to only occur when the editor has focus.

This PR fixes issue #2317.


sbauer322:
Added benefit of this PR is that now 1 - 9 on the numpad work too... so `Ctrl-1` will go to the first tab, for example.

sbauer322:
I will be merging this after February 25th, barring any comments or concerns.

PR #2321, opened 2017-02-03
Allow build and start on MINGW (no cygwin required)

domoran:
Disclaimer: I read in the contribution guidelines, that you do not want contributions to "scripts" ... 

Anyway I thought I would add that a windows build does also work without cygwin installed (which is kind of a 'heavy' environment). 

I found that an installed GIT is sufficient to build LightTable, since it includes a "bash" shell. Therefore I added inside the build and start scripts the check for the MINGW environment and put another chapter in the documentation for people, that do not want to install cygwin.

### DIFF
--- a/doc/developer-install.md
+++ b/doc/developer-install.md
@@ -11,9 +11,10 @@
 
 None. Skip to [Build](#build).
 
-### Windows
 
-Download [Cygwin](https://cygwin.com/install.html).
+### Windows - with Cygwin
+
+Download [Cygwin](https://cygwin.com/install.html). If you find Cygwin to much on your windows system, take a look at [building with Windows (using GIT)](#WindowsGit)
 
 You may need to change */etc/fstab* (path in Cygwin) to fix a directory permission error when building. Change this line:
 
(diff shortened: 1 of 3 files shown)


sbauer322:
I'll be able to give it a try in within the next week or so, but it sounds promising!

@LightTable/maintainers - if possible, I would prefer to have multiple people test this out prior to merging.

sbauer322:
I will have access to a Windows 8.1 machine tomorrow night to test this.

sbauer322:
Seems I run into a problem with an error when attempting to run `script/build.sh` with this PR:

```
script/build.sh: line 25: node_modules/.bin/grunt: No such file or directory
```

Briefly poking at line 25 and then looking in the contents of `deploy/electron/node_modules/.bin/` suggests that there is indeed no file named "grunt"... there is a "grunt.cmd" though.

Aside from the error above, there are a couple of other minor things:

- Could you make it more even more obvious that you would need to use the lein _script_ and not the .bat... and also that the script must be in the path (even if the bat is on the path). I managed to overlook the difference and got stuck on it not working with the .bat for a bit.
- It seems that part of the changes to `build-app.sh` and `light.sh` is that every line now ends in a whitespace. Could you fix this?

Overall, I think it just has a couple problems but should be in a good state soon enough. Thank you @domoran!

Issue #2178, opened 2016-04-05
Select text in Plugins


1j01:
You should be able to select text in the Plugins view.
The text "by" shouldn't be a psuedo element, and there should probably be an actual space between the package name and version, but other than that it just needs `-webkit-user-select: text`.

kenny-evitt:
Interesting. This seems to be explicit; [the relevant CSS](https://github.com/LightTable/LightTable/blob/53a90bee4c2c6758e4b5dffdffbc32256a05e00d/deploy/core/css/structure.css#L14):

```
body { -webkit-user-select: none; }
```

@cldwalker @rundis Any ideas about whether removing this CSS rule is a bad idea?

I commented the line out via Developer Tools and I'll update this issue if I run into anything that seems related to that change.

kenny-evitt:
@1j01 Why should "by" text not be a pseudo-element?

1j01:
We should keep the global `-webkit-user-select: none;` and add` -webkit-user-select: text;` to `.plugin-manager li`, and we should avoid textual pseudo elements so that when you copy and paste, the text is present. It wouldn't be the biggest issue because it's still fairly clear the second line is the author:
```
Paredit0.0.4
Kodowa
Paredit integration for Light Table
```
But it would be better if it copied it as it appeared.

kenny-evitt:
@1j01 How is keeping the global select-none helpful? And how is it helpful to be able to copy "by"? I'm not against making the changes, but I don't know why they were implemented or implemented that way nor what consequences changing them would have other than being able to copy the plugin text. And I'm not sure of the utility of being able to copy the plugin text either.

1j01:
@kenny-evitt I'm was just guessing it would break stuff to remove it. Trying it out, it's not too bad (and it's good to be able to select as much as possible), but it would probably be better to enable it specifically for each UI element (like the line / column view, the individual file/folder list items (although it would need to ignore clicks when there's selected text to be able to select the name of a file or folder easily))

Like I was trying to say, "by" wouldn't matter too much, but there's not really any point in having it be a psuedo element, so it might as well be copyable with the rest of the text

kenny-evitt:
@1j01 I have no strong feelings about this so if you'd like the changes to be made, please submit a pull request and we'll review and test your proposed changes.

sbauer322:
For anyone curious, below is part of the Plugin Manager where text cannot be selected;

[...]

PR #2322, opened 2017-02-07
Select text in Plugin Manager

ahoffer:
Enhancement to select text in Plugin Manager view
  See issue https://github.com/LightTable/LightTable/issues/2178

Also modified .gitignore file to ignore IntelliJ's internal files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,5 @@
 .lein-plugins/
 .nrepl-port
 /builds/
+.idea
+*.iml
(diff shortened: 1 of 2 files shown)


sbauer322:
Tested it out and it works great, thanks!

I did notice that the underline for plugin titles was removed, but I think I prefer it without this underline since the cursor still changes to indicate it is a link and there is less visual clutter.

@LightTable/maintainers - I am going to leave this open until at least February 11th in order to give anyone interested a chance to review.

ahoffer:
I removed the underline for plugin titles because it looked cluttered.

sbauer322:
I think it was a good call. Thank you for the PR, @ahoffer!

Issue #2313, opened 2017-01-16
Better portability: build on Windows + Git Bash to run script/light.sh


robjens:
Currently, the `script/light.sh` file ran after building this solution on a Windows machine, doesn't pick up the fact that I can actually run LightTable just fine.

At the moment, it checks first for Darwin - which I'd keep since it to has bash - but then proceeds to check `Linux` using `uname -s` as well, finally to check for Cygwin in the same fashion.

Now, I am not sure if and why Cygwin would need to run the `.exe` executable but I presume this to be correct. However, in my case and sense, the following would have been a more reliable way to check:

```sh
 hash bash 2>/dev/null && {
    CLI="${DIR}/deploy/electron/electron/electron"
}
```

If Cygwin does in fact require the executable and somehow is incapable of running the shell file, it should come first. But the code above would run on any bash capable/installed shell in Windows.

For the record, my Git Bash reports `MINGW64_NT-10.0` which makes me inclined to think that whole string isn't really reliable but I guess one could work with the `MINGW` bit.

I wouldn't mind doing a PR but figured I should ask first if this would be a desirable change. The used snippet is the most idiomatic and portable way of checking if applications are installed in bash - it should work on literally any *nix that has it.

sbauer322:
I'm not opposed to a PR for this... As long as everything still works! :smile: 

@LightTable/maintainers - any concerns or objections?

Mouvedia:
@robjens while you are at it could you check bash on <abbr title="Windows Subsystem for Linux">WSL</abbr>?

rundis:
No objections, as long as it works... and we are able to test that it does I'm good.

sbauer322:
Sounds like a PR for this would be appreciated @robjens - feel free to proceed if you so desire.

robjens:
Alright, will do... cheers. I'm not sure if I'm supposed to sign a CA but I'll check. Also I noticed it didn't work out of the box as expected when I changed the script pre-build but it might have also overwritten something so I'll take care to double check before I do a PR.

sbauer322:
No need for a CA from us!

robjens:
@sbauer322 Cool. I just built LT again on Windows and I changed the light.sh script before running build.sh for the first time. On compilation/setup finish, it tries to run script/light.sh at the very end right? It seems it may not be using the local one though, I'm guessing it uses the git cloned script ...? Since it basically notifies right after the git activity...

[...]

PR #2323, opened 2017-02-23
#2313 More portable nix/bash detection to include mingw

robjens:
Mingw doesn't seem to have `expr` in its environment so the script failed to detect when run in e.g. Git Bash on Windows. This would fix it and should work on any Linux at the same time. I installed Cygwin just to ensure that the generic second condition wouldn't override Cygwin last, it doesn't.

### DIFF
--- a/script/light.sh
+++ b/script/light.sh
@@ -10,7 +10,7 @@ DIR=$(pwd)
 
 if [ "$(uname)" == "Darwin" ]; then
   CLI="${DIR}/deploy/electron/electron/Electron.app/Contents/MacOS/Electron"
-elif [ "$(expr substr $(uname -s) 1 5)" == "Linux" ]; then
+elif hash bash 2>/dev/null; then
   CLI="${DIR}/deploy/electron/electron/electron"
 elif [ "$(expr substr $(uname -s) 1 9)" == "CYGWIN_NT" ]; then
   CLI="${DIR}/deploy/electron/electron/electron.exe"


sbauer322:
This looks useful @robjens - thank you!

I should be able to test this tomorrow, but just a few observations in the meantime:
- Does the line changed mean that bash will now be required when building on Linux? I understand bash is fairly ubiquitous on Linux, but some people might have replaced it for something else.
- Would you mind throwing in a comment on the line to indicate that it is for mingw / Git Bash on Windows / Linux (assuming the above isn't an issue)? It won't be obvious to most why the line is different from the others.
- Assuming everything works out, could you also update the [developer-install](https://github.com/LightTable/LightTable/blob/master/doc/developer-install.md) in this PR since Cygwin will no longer be the only option for Windows?

Sorry for all the questions! 😄

sbauer322:
I seem to be experiencing the same problem as described here https://github.com/LightTable/LightTable/pull/2321#issuecomment-280548998 .

Would you be able to test what happens when you run `script/build.sh` from MINGW? I am currently testing this with GitBash and keep getting the `script/build.sh: line 25: node_modules/.bin/grunt: No such file or directory` message.

Do you have `grunt` or `grunt.cmd` in your `deploy\electron\node_modules\.bin` directory (or both)?

sbauer322:
This seems to work just fine on Windows with GitBash... despite my initial unrelated caching problem.

For Linux, as expected, if you don't have bash installed then the script won't work. However, since the script already requires bash prior to this PR I no longer feel this is an issue for this PR.

If you could make the changes mentioned in an earlier comment I think this will be fine, @robjens.

@LightTable/maintainers anyone available such that we can get another person to QA this?

domoran:
It seems we had the same idea on this pull request [#2321](https://github.com/LightTable/LightTable/pull/2321). 

I already made the requested adaptions to the developer docs.  How shall we proceed here?

sbauer322:
It is certainly a rare case to have multiple PRs addressing the same thing here. I don't have a preference for which one is used as long as it addresses the concerns mentioned.

It might be simplest to reduce @domoran's PR to just the developer docs and merge it at the same time as @robjens's PR is merged?

If one contributor or the other is too busy to work on their PR then we can simply take the other PR instead.

sbauer322:
I've been using the fix from this PR and #2321 on Windows 8.1 and have not ran into any troubles.

Issue #2030, opened 2015-10-29
LightTable.exe file for 0.8.0-alpha doesn't show LT icon on Windows (7), and related issues


kenny-evitt:
The file shows the Electron icon instead.

I also noticed on my Windows 7 netbook that the window title, when it first opens after running the executable, shows "Electron". My primary Windows 7 computer is much faster so I didn't notice this until I looked for it just now.

I also noticed that the executable metadata includes GitHub as the publisher (?) when granting permission for network access via Windows Firewall.

cldwalker:
@kenny-evitt I took the beginner label off this since the label is linked from CONTRIBUTING and should have links in the code and possible solutions to make the contributing approachable for a beginner

cldwalker:
@kenny-evitt Apologies if this wasn't obvious. Updated description [of beginner label](https://github.com/LightTable/LightTable/wiki/For-Contributors#issue-labels)

kenny-evitt:
@cldwalker No worries; thanks for clarifying.

TimothyDJewell:
Hey guys, I noticed this is still open when I installed 0.8.1 on my Windows 10 machine.  Unfortunately, this bug makes an otherwise shiny IDE seem a little...confused.

I was looking into solutions, and I found a note in electron-packager about [special handling necessary for building Windows exes](https://github.com/electron-userland/electron-packager#building-windows-apps-from-non-windows-platforms).  My understanding is that we would need to manually edit the exe to make it display the correct information.  I'm not familiar with electron or this project, so I'm not sure whether this is a normal or acceptable thing for this project to handle.

kenny-evitt:
@TimothyDJewell We build the Windows release on a Windows computer so the special handling mentioned in the link you referenced isn't an issue for us. We also have a Windows-based CI build (currently just for a development branch) too.

TimothyDJewell:
@kenny-evitt I think their comment about needing to edit the exe  and how they do that applies to all platforms as context to why they need special handling for non-Windows platforms. I don't think the root of this issue is in the current build process, per se, so much as that Windows requires an extra build step of editing the exe so that the icon and metadata are accurate. I don't know if Windows has another way for this to work.

To ask a more direct question: Would the project maintainers be receptive to additional Windows build dependencies to resolve this issue? (Based on what I'very seen, it seems to point to rcedit or the node interface for it.) If so, is there a preferred way for this to be done?
If I'm totally misunderstanding the issue here, don't hesitate to correct.

[...]

PR #2324, opened 2017-02-26
Rebrand built exe for Windows

TimothyDJewell:
By default it has GitHub's information, as reported in #2030.  I know the contributing guidelines say "core team only" for scripts/, but this change is quite straightforward and not directly related to the actual build process.  For me, this has served as a distraction every time I go to use LightTable on my Windows machine (seeing Electron's icon and branding), so I'd like to just fix it.  I'm not that familiar with cygwin/this project's processes, so please don't hesitate to point out where I'm wrong.  I'll do my best to fix things up however I can.

I built the .ico file in GIMP from the existing lticon.png using the instructions described [on StackOverflow](http://stackoverflow.com/a/10819673).  The requirement of this PR should be simply that [rcedit.exe](https://github.com/electron/rcedit) be located somewhere on the build machine's path.  If it's not there, it should just fall back to the existing behavior, but with a console output from which and echo indicating the failure to rebrand.

Other relevant links:
 - [The electron team's notes about how to rebrand for Windows](http://electron.atom.io/docs/tutorial/application-distribution/#rebranding-with-downloaded-binaries)
 - [grunt-rcedit](https://github.com/unindented/grunt-rcedit#version-information) had the most helpful information about which version strings, though it points back to MSDN.


sbauer322:
Thanks for this @TimothyDJewell, I will be able to give it a try later this week.

If anyone else wishes to review it then feel free to do so!

sbauer322:
Gave this a try and it mostly works! Pretty nice to see the Light Table icon on Windows. Thank you @TimothyDJewell!

Just a couple things:

- Could you update `doc/developer-install`'s Window's section to specify that rcedit will be necessary. Also, please mention that there should be a release of rcedit available at https://github.com/electron/rcedit/releases ... Their README doesn't indicate that there is a release (as opposed to building it), which I imagine will be misleading to someone at some point.
- The Light Table icon is not displayed for certain views. They are 'small', 'list', and 'details'. Works for all the rest though (including as a shortcut on the desktop). Could you address this, please?

Overall, this looks great.

TimothyDJewell:
Sure, I'll get on the documentation.  With the icon, I ran into a similar issue while testing this, but it turned out that my problem was Windows icon cache rather than the icon itself.  If you [refresh the cache](http://superuser.com/questions/499078/refresh-icon-cache-without-rebooting), does it get corrected?

sbauer322:
Heh, you are correct. It was a caching issue. Just updating the documentation should be fine.

TimothyDJewell:
@sbauer322 I've added it to the docs, but I'm not sure I'm happy with the way it reads.  I'm certainly open to feedback if there's a preferred way to do this.

sbauer322:
How about something like the following:

In order to rebrand a build's resulting executable, you will need rcedit on your path to allow the build to brand the executable with the Light Table icon.... Otherwise, it will be branded with the default electron icon. There is a [pre-built release](https://github.com/electron/rcedit/releases) of rcedit available on GitHub.

sbauer322:
Looks great @TimothyDJewell, thank you!

PR #2329, opened 2017-04-12
toggle-find-bar

ahoffer:
- Added command, key mapping, and behavior to toggle find bar's visibility.
- Partial solution for issue 2285.
- I still don't know much about how LightTable works. Submitting this PR to start constructive conversation about how this feature should be implemented. 
- Once functionality for the find bar looks good, I'll tackle search (per the github issue)


sbauer322:
Everything should be under `src/objs/find.cljs`. Since the bar already implements showing and hiding, perhaps it would be easiest to create something like a `:find-toggle` command that simply calls `:find.show` and `:find.hide` as appropriate? If we are fortunate then the `:find-bar` object might even already be correctly keeping track of its `:shown` field... which would make creating the toggle command easier.

Some relevant commands involved are: `:find.show` and `:find.hide`

Some relevant behaviors involved are `:show!`, and `:hide!`.

Once `:find-toggle` is working, you would then update the `deploy/settings/default/default.keymap` such that the "pmeta-f" entry is something like `[:editor.keys.normal "pmeta-f" :find.fill-selection :find.show]`.

That seems like the most painless solution after spending a few minutes looking at the code... of course, there could be better solutions.

Since we are working on improving our documentation, anything that is unclear to you would be useful to know about... our [Gitter](https://gitter.im/LightTable/LightTable) would be a good place for that. Let me know if you run into any troubles, or need more information (especially with how LT is laid out).

ahoffer:
@sbauer322. Looks like a messed up the commit. I don't know why there are no changes in it. Let me fix that.

Issue #2285, opened 2016-10-16
Find and search cannot be toggled off


sbauer322:
Find (Ctrl-F) and search (Ctrl-Shift-F) cannot be toggled off by the same command that opened it. This is inconsistent from other temporary widgets such as language docs (Ctrl-Shift-D), command pane (Ctrl-Space), and the console. 

Find can only be hidden via the command pane. Search can be closed by closing its tab. For consistency,  these should be toggle-able.

Mouvedia:
I dunno about the desired behaviour from the user point of view in that scenario:
* search something which exists in the editor
* focus search field
* esc
* pmeta-f (`:find.show`)

I was surprised that the queried keyword wasn't automatically highlighted in the editor. Is this expected?

ahoffer:
@sbauer322 
I am not sure what behavior ctrl+shift+f should exhibit. Ctrl_shift+f open the Searcher, which is a new tab. The new tab is made active. 

Should ctrl+shift+f close the Searcher tab and set focus to the tab on the left? This is what happens when you close the Searcher tab by clicking on the tab's "X" button.

Personally, I expected subsequent ctrl+shift+f invocations to open more Searchers tabs. Why limit users to one? I might want a Search tab for all references to "object/merge" strings and another Searcher tab for "<div>" strings.

What behavior satisfies the "principle of least surprise" for multiple ctrl+shift+f ?

sbauer322:
Hmm, those are some good questions @ahoffer! 

There are a couple different paths to pursue, but in hindsight, this issue should not have included the `Ctrl-Shift-F` Search. As you rightly pointed out, since the tab has an "X" button, it is not necessary to be able to close the Search tab via `Ctrl-Shift-F`... additionally, whatever keybinding already exists to close a tab should be able to handle closing the Search tab already. This combination sufficiently handles closing the Search.

I am leaning towards narrowing the scope of this issue to just the `Ctrl-F` find bar and merge the related PR, if you don't mind. Another issue could then be opened for the Search to handle multiple Search independent results. Let me know how you want to proceed, @ahoffer.

---------------

(Just tossing this here since we are on the topic of Search and it can be used as the basis of the new Search issue mentioned above)

Regarding what the behavior should be for the Search, I can't say I have the right answer, but let's pick something and give it a try. Worst case we just change/revert the behavior. That said, I do not know offhand how difficult it would be to implement some of those features (especially additional search tabs).

The following are a few possible paths, but I am open to alternatives.

[...]

PR #2330, opened 2017-04-13
Toggle find bar

ahoffer:
Partial solution to issue https://github.com/LightTable/LightTable/issues/2285-- shows/hides find bar, does not yet address search.
The implementation is pretty similar to @sbauer322's notes in a previous PR (that PR is since closed). However, Scott's notes implied it is possible for one command to directly invoke another command. If that's the case, this implementation could be simplified because no additional behavior would be necessary. If I can find an example of how to do that, I'll update this PR with another commit.

Two other comments. First, I didn't make any progress on this issue until I came across this tutorial: http://blog.jakubarnold.cz/2014/01/13/light-table-plugin-tutorial.html. Once I got the sample code in the tutorial working, it was clear how I could define and add a behavior. Until then, I didn't grok the use of the behavior-object-tag paradigm. Second, I am indebted to this page: https://github.com/shaunlebron/How-To-Debug-CLJS. Without it, I would not have made any headway.


ahoffer:
UPDATE: I got it working calling the find and show commands directly; no need for a "toggle" behavior. However, I still needed to set the :shown properties of the find bar.

sbauer322:
Seems to work quite nicely. If you could squash the commits into a single commit, I think it would be perfect.

Thanks for the good work, @ahoffer!

@LightTable/maintainers - I will merge this after 2017-04-22 unless there is additional feedback to address.

ahoffer:
@sbauer322 If it looks good, I go ahead and do the same for the search side of things.

sbauer322:
Sounds good! They do have a squash on merge, so I'll just use that to make life easier.

sbauer322:
@LightTable/maintainers - This PR is ready for review. Discussion in issue #2285 has narrowed the scope of work to just the Find bar. I will be merging after 2017-04-29 unless there is feedback or concerns.

rundis:
Looks good. This has annoyed me for ages, but apperantly i´ve gotten used to it. Tx  !

PR #2337, opened 2017-05-10
Fix Control Pane's arrow direction

### LINKED ISSUE #2286: Connection Pane's arrow is the same in both states
As the title says, the Connection Pane's arrow is always in the same direction.

Here is the initial state, where current connections are displayed:

connection-pane-arrow-incorrect

This arrow likely should be pointing downward.

For comparison, when you click on the 'Add Connection' button, the arrow is in the correct position indicating that it can be collapsed:

connection-pane-arrow-correct

Alternatively, the arrow could simply be removed.

pndpo:
Resolves #2286 with a single line added to structure.css.


sbauer322:
Thanks for the fix!

Issue #2342, opened 2017-06-22
Dragging a tab outside of Light Table produces incorrect text


sbauer322:
As seen [here](https://t.co/saY4lyaUzb), the string "dummy" is used. The expected behavior would be the path of the document or just the document name.

Confirmed the behavior on Windows 8.1 when dragging to FireFox and Chrome. Dragging appears disabled when dragging to other applications like Notepad and Notepad++.

sbauer322:
PR merged.

PR #2355, opened 2017-10-16
Remove unused dataTransfer code from dragdrop.js

NGaffney:
Some scaffolding around drag and drop dataTransfer objects was causing LightTable/LightTable#2342

I've just taken the problematic code out and not replaced it with anything.


sbauer322:
Tried this out and it does prevent the tabs drag and drop producing 'dummy'.

sbauer322:
Thanks for the PR!

Issue #2336, opened 2017-04-24
Small UI tweaks - `font-family` and placeholder text


pedrorgirardi:
I noticed that the `find bar` and the command palette `search` text looks different from the other parts of LT:

`find bar`

![screen shot 2017-04-24 at 16 12 35](https://cloud.githubusercontent.com/assets/585191/25355340/b03bac80-290c-11e7-9c31-0a45e574f3e9.png)

`command palette`

![screen shot 2017-04-24 at 16 16 43](https://cloud.githubusercontent.com/assets/585191/25355377/ca3d0ab6-290c-11e7-8a79-7345f96c8392.png)

LT uses `Verdana`, if `Verdana` is "applied" it would look like:

`find bar`

![screen shot 2017-04-24 at 16 14 12](https://cloud.githubusercontent.com/assets/585191/25355462/24585384-290d-11e7-8aea-a51451603551.png)

or (with the first letter capitalized)

![screen shot 2017-04-24 at 16 14 48](https://cloud.githubusercontent.com/assets/585191/25355499/45f12d4a-290d-11e7-80d9-4670d52a32f7.png)

`command palette`

![screen shot 2017-04-24 at 16 14 56](https://cloud.githubusercontent.com/assets/585191/25355521/51fdfc08-290d-11e7-9e83-44f562f7dc18.png)

or (with the first letter capitalized)

![screen shot 2017-04-24 at 16 15 18](https://cloud.githubusercontent.com/assets/585191/25355533/59540d58-290d-11e7-8594-09e2a9d822a4.png)

Is that intentional or a "mistake"? It it's not intentional I can create a PR with the CSS changes - if that's okay for you. 

Thanks guys!

sbauer322:
I'm not aware of this being intention, but I would be open to a pull request correcting both the font and capitalization. I believe the first letter should be capitalized as it is consistent with other parts of LT.

Thanks for noticing this!

pedrorgirardi:
Cool! I will submit a PR then.

Thanks, @sbauer322!

denis631:
Finally made a PR for this. This was annoying me for a long time

PR #2372, opened 2018-05-30
[#2336] set Verdana family to be default for search input.

denis631:
Console FontFamily without quotes fix.
Welcome screen text, button labels are now of Inconsolata/UbuntuMono family


denis631:
I guess I am not doing this the right way. Overriding all this doesn't feel right to me. Any ideas why this is happening? In the structure.css default font for body elements is set, but is not used. What do you think?

